PR #5, opened 2017-06-15
Adding Ansible Shared specs from ManageIQ repo

juliancheal:
We missed this folder during extraction original files from
https://github.com/ManageIQ/manageiq/tree/master/spec/support/ansible_shared


juliancheal:
> ❗️ - Line 25, Col 48 - Style/AlignHash - Align the elements of a hash literal if they span more than one line.

In this instance when you align the hashes it makes the code look really unreadable.

PR #4, opened 2017-06-07
Add missing VCR Cassettes

juliancheal:

PR #6, opened 2017-06-15
Load shared examples and vcr cassettes from self

jameswnl:
## Before
Loading shared examples and vcr cassettes from locally cloned `manageiq` (which are already moved to current repo and about to be removed from `manageiq`.
Refer to merged PRS: #4, #5 

Without this PR, once these are removed rom `manageiq` repo, the specs will fail

## After
They are loaded from current repo.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -4,9 +4,9 @@ if ENV['CI']
 end
 
 # Uncomment in case you use vcr cassettes
-#VCR.configure do |config|
-#  config.ignore_hosts 'codeclimate.com' if ENV['CI']
-#  config.cassette_library_dir = File.join(ManageIQ::Providers::AnsibleTower::Engine.root, 'spec/vcr_cassettes')
-#end
+VCR.configure do |config|
+  config.ignore_hosts 'codeclimate.com' if ENV['CI']
+  config.cassette_library_dir = File.join(ManageIQ::Providers::AnsibleTower::Engine.root, 'spec/vcr_cassettes')
+end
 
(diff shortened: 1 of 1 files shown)


jameswnl:
@miq-bot remove_label wip

jameswnl:
@miq-bot add_labels developer

Fryguy:
If your statement is true, then this would fail for *all* providers, right?

jameswnl:
@Fryguy if other provider repos have shared examples and have moved their portion in `spec/shared/` to their own repo.

They may still be passing by loading the cloned `manageiq/spec/*`.

(I am assuming there's nothing in my local env fixing the `Rails.root` to `manage`.)

jameswnl:
@miq-bot add_label fine/no

juliancheal:
LGTM 👍

Fryguy:
I see this line in *every* provider
https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/spec/spec_helper.rb#L11
https://github.com/ManageIQ/manageiq-providers-azure/blob/master/spec/spec_helper.rb#L12
https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/spec/spec_helper.rb#L11
...

jameswnl:
@Fryguy I've created a [PR to fix the generator template](https://github.com/ManageIQ/manageiq/pull/15416/files). 

So for existing providers, do we update the generator code and then rerun the generator to fix this problem? Or this PR (plus the other provider repos) can go in?

Fryguy:
@durandom I'll merge this for now to unstick @jameswnl , but I still question if it's being done correctly at the generator level in https://github.com/ManageIQ/manageiq/pull/15416 .  Whatever change we decide over there will have to work for this repo anyway and will probably require PRs to all the repos to update for consistency.

PR #9, opened 2017-06-27
New Tower credential types

jameswnl:
Make MIQ's support to be on par with what Tower supports.

https://www.pivotaltracker.com/story/show/148234657

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/shared/automation_manager/openstack_credential.rb
+++ b/app/models/manageiq/providers/ansible_tower/shared/automation_manager/openstack_credential.rb
@@ -4,13 +4,13 @@ module ManageIQ::Providers::AnsibleTower::Shared::AutomationManager::OpenstackCr
   COMMON_ATTRIBUTES = {
     :userid => {
       :label     => N_('Username'),
-      :help_text => N_('Username for this credential'),
+      :help_text => N_('The username to use to connect to OpenStack'),
       :required  => true
     },
     :password => {
       :type      => :password,
       :label     => N_('Password (API Key)'),
-      :help_text => N_('Password for this credential'),
+      :help_text => N_('The password or API key to use to connect to OpenStack'),
       :required  => true
     }
   }.freeze
--- a/app/models/manageiq/providers/ansible_tower/shared/automation_manager/openstack_credential.rb
+++ b/app/models/manageiq/providers/ansible_tower/shared/automation_manager/openstack_credential.rb
@@ -1,0 +1,49 @@
+module ManageIQ::Providers::AnsibleTower::Shared::AutomationManager::OpenstackCredential
+  extend ActiveSupport::Concern
+
+  COMMON_ATTRIBUTES = {
+    :userid => {
+      :label     => N_('Username'),
+      :help_text => N_('Username for this credential'),
+      :required  => true
+    },
+    :password => {
+      :type      => :password,
+      :label     => N_('Password (API Key)'),
+      :help_text => N_('Password for this credential'),
+      :required  => true
+    }
+  }.freeze
+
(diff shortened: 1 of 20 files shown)


jameswnl:
@miq-bot remove_label wip

bronaghs:
hey @durandom - Can you please view this? Danke!

app/models/manageiq/providers/ansible_tower/shared/automation_manager/openstack_credential.rb
```diff
@@ -0,0 +1,49 @@
+module ManageIQ::Providers::AnsibleTower::Shared::AutomationManager::OpenstackCredential
+  extend ActiveSupport::Concern
```
durandom: I see you are following the pattern in the other credential modules, but why is `AS::Concern` used here?

lib/tasks_private/spec_helper.rake
```diff
@@ -80,6 +80,37 @@ class PopulateTower
   end
 
   def create_dataset
+
+    ssh_key_data = <<~HEREDOC
+    -----BEGIN RSA PRIVATE KEY-----
```
durandom: thanks for sharing you private ssh key 😈

spec/support/ansible_shared/automation_manager/refresher.rb
```diff
@@ -100,12 +100,12 @@
   def assert_counts
     expect(Provider.count).to                                 eq(1)
     expect(automation_manager).to                             have_attributes(:api_version => "3.0.1")
-    expect(automation_manager.configured_systems.count).to    eq(130)
```
durandom: these counts are a pain.
in aws we changed this to get the counts straight from the API and just assert they are equal to objects counts in the DB.

You would access the API easily via `automation_manager.provider.connect.hosts.count`
Maybe add a safe trap, that the numbers are non-zero

lib/tasks_private/spec_helper.rake
jameswnl: you are welcome. generated for this particular purpose only 😄

spec/support/ansible_shared/automation_manager/refresher.rb
jameswnl: yeah, was actually considering this. but thinking of using the `Faraday` direct call to avoid depending on the `Ansible Tower Gem`. What do you think? 
`automation_manager.provider.connect.hosts.count` would be much more straightforward.

jameswnl: @durandom let's keep this as is for now. There's some strange issue with the Tower we have been using to perform this test.

(The reported count of hosts is more than the actual hosts returned)

app/models/manageiq/providers/ansible_tower/shared/automation_manager/openstack_credential.rb
jameswnl: Done!

spec/support/ansible_shared/automation_manager/refresher.rb
durandom: > The reported count of hosts is more than the actual hosts returned

🤔 

If you have some spare time 😆 maybe you want to add a `docker run` for setting up a tower and populating it?

jameswnl: I did and ended up aborting it because it needs enterprise license to support survey spec.

jameswnl:
@miq-bot add_labels fine/yes

simaishi:
@jameswnl Is there a BZ for this?  Can you please create if it doesn't exist?

simaishi:

PR #12, opened 2017-07-14
Fix caps for network credential type

mzazrivec:
The string is propagated into API & UI, therefore the use of capitals here should be consistent
with all the other credential types.


jameswnl:
LGTM

PR #15, opened 2017-07-19
Ensure shared example groups use shared resources

bdunne:
Most of this is whitespace change caused by the new block


bdunne:
@Fryguy Please review

PR #13, opened 2017-07-14
Azure Classic credential added

jameswnl:
It is deprecated by Ansible Tower. But PM seems think this can still be useful to support it.

@miq-bot add_labels enhancement

PR #16, opened 2017-07-21
Specs added for azure-classic-credential

jameswnl:
Previously didn't add this to #13 because it will break `manageiq` `embedded_ansible`.  

This should follow https://github.com/ManageIQ/manageiq/pull/15626


bronaghs:
LGTM

simaishi:

PR #19, opened 2017-08-08
adding require_nested for new azure_classic_credential

jameswnl:
@miq-bot add_labels bug


simaishi:

PR #20, opened 2017-08-08
validate provider/manager before queuing create_in_provider

jameswnl:
https://bugzilla.redhat.com/show_bug.cgi?id=1476327

@miq-bot add_label bug


jameswnl:
@miq-bot assign @blomquisg

jameswnl:
@blomquisg can you take a look?

jameswnl:
@bdunne can you review this?

app/models/manageiq/providers/ansible_tower/shared/automation_manager/tower_api.rb
```diff
@@ -21,6 +23,9 @@ def create_in_provider(manager_id, params)
     def create_in_provider_queue(manager_id, params, auth_user = nil)
       process_secrets(params) if respond_to?(:process_secrets)
       manager = ExtManagementSystem.find(manager_id)
+      if ManageIQ::Providers::Inflector.provider_name(manager) != ManageIQ::Providers::Inflector.provider_name(self)
```
bdunne: It feels strange that we've gotten this far without an error.  I know that we want the create to happen in the provider first and then refresh to bring the record into our database, but should this live as a validation on the model instead?  @Fryguy Any thoughts on where this belongs?

app/models/manageiq/providers/ansible_tower/shared/automation_manager/tower_api.rb
```diff
@@ -1,6 +1,8 @@
 module ManageIQ::Providers::AnsibleTower::Shared::AutomationManager::TowerApi
   extend ActiveSupport::Concern
 
+  class UnsupportedManagerOperation < StandardError; end
```
bdunne: I would expect all ManageIQ error classes to be a `MiqException` and to be defined here: https://github.com/ManageIQ/manageiq-gems-pending/blob/master/lib/gems/pending/util/miq-exception.rb

app/models/manageiq/providers/ansible_tower/shared/automation_manager/tower_api.rb
jameswnl: This only happens through calls to API since calls through API allow user to provide arbitrary `manager_id`

bdunne: Right, but adding an Ansible credential type to a Foreman provider doesn't make sense on either system, that's why I was suggesting the validation on the model.

Fryguy: The problem here is that it is doing an ExtManagementSystem.find instead of the more targeted find of only valid EMSes.

spec/support/ansible_shared/automation_manager/credential.rb
```diff
@@ -79,6 +79,11 @@
       )
     end
 
+    it ".create_in_provider_queue to fail with incompatible manager" do
+      wrong_manager = FactoryGirl.create(:provider_foreman).managers.first
```
bdunne: You only need a single manager that is the wrong type, not a Provider, Manager and Authentication.  Avoid creating extra database records and looking them up.  Prefer the `:ext_management_system` or `:automation_manager` factory.

jameswnl: Right! Done

app/models/manageiq/providers/ansible_tower/shared/automation_manager/tower_api.rb
```diff
@@ -24,7 +24,7 @@ def create_in_provider(manager_id, params)
 
     def create_in_provider_queue(manager_id, params, auth_user = nil)
       process_secrets(params) if respond_to?(:process_secrets)
-      manager = ExtManagementSystem.find(manager_id)
+      manager = self.parent.find(manager_id)
```
bdunne: Looks like you're just down to the rubocop comment about `self`

PR #25, opened 2017-10-11
Split #create_in_provider method into two methods.

bzwei:
1. Extract api raw creation logic into a separate method. It is needed if we don't want a refresh.
2. Do not associate a job with the job template if it is not persisted. In our use case to run a playbook using a temperate job template we do not persist or refresh the job template.

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/shared/automation_manager/job.rb
+++ b/app/models/manageiq/providers/ansible_tower/shared/automation_manager/job.rb
@@ -9,9 +9,10 @@ module ManageIQ::Providers::AnsibleTower::Shared::AutomationManager::Job
     #   :extra_vars => Hash
     #
     def create_stack(template, options = {})
-      self.new(:name                  => template.name,
+      template_ref = template.new_record? ? nil : template
+      new(:name                  => template.name,
           :ext_management_system => template.manager,
-          :job_template          => template).tap do |stack|
+          :job_template          => template_ref).tap do |stack|
         stack.send(:update_with_provider_object, raw_create_stack(template, options))
       end
     end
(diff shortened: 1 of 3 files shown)


bzwei:
@miq-bot assign @gmcculloug 

app/models/manageiq/providers/ansible_tower/shared/automation_manager/job.rb
```diff
@@ -9,9 +9,10 @@ module ClassMethods
     #   :extra_vars => Hash
     #
     def create_stack(template, options = {})
-      self.new(:name                  => template.name,
+      template_ref = template.new_record? ? nil : template
+      new(:name                  => template.name,
           :ext_management_system => template.manager,
-          :job_template          => template).tap do |stack|
+          :job_template          => template_ref).tap do |stack|
```
syncrou: @bzwei - Setting this to nil isn't going to cause any other surprises?

It seems odd that we'll have a AnsibleTowerJob without a referenced JobTemplate (ConfigurationScript.) It almost feels like this should be a different type of subclassed Job since it will never have what a normal AnsibleTowerJob would have - a referenced JobTemplate.

In that case we could override create_stack specifically for the runner.

bzwei: So we don't have the use case that expects the template from a job (stack) can never be `nil`. For Azure provider it is actually common that we cannot get the template for a stack.

We could create a subclass of Job class to separate this use from normal Ansible jobs, but I feel it is not very helpful. When we query tower jobs the subclass instances are included anyway.

app/models/manageiq/providers/ansible_tower/shared/automation_manager/tower_api.rb
```diff
@@ -2,11 +2,15 @@ module ManageIQ::Providers::AnsibleTower::Shared::AutomationManager::TowerApi
   extend ActiveSupport::Concern
 
   module ClassMethods
-    def create_in_provider(manager_id, params)
+    def create_in_provider_raw(manager, params)
```
bdunne: We usually name these methods `raw_something`

PR #27, opened 2017-10-18
Extract method raw_delete_in_provider from delete_in_provider

bzwei:
Same reason as #25, now extract `#raw_delete_in_provider` from `#delete_in_provider`


bzwei:
@miq-bot assign @gmcculloug

PR #34, opened 2017-11-06
Correct extra_var keys to original case

bzwei:
When the extra_vars are passed through automate as described in the bugzilla, keys are converted to lower case. The original definition of keys may be pulled back from template's `extra_vars` and `survey_spec`. Note that the fix is not able to find the original key definition if it is not in either field. 

https://bugzilla.redhat.com/show_bug.cgi?id=1508946

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/shared/automation_manager/job.rb
+++ b/app/models/manageiq/providers/ansible_tower/shared/automation_manager/job.rb
@@ -21,7 +21,7 @@ module ManageIQ::Providers::AnsibleTower::Shared::AutomationManager::Job
       options = reconcile_extra_vars_keys(template, options)
       template.run(options)
     rescue => err
-      _log.error "Failed to create job from template(#{name}), error: #{err}"
+      _log.error("Failed to create job from template(#{template.name}), error: #{err}")
       raise MiqException::MiqOrchestrationProvisionError, err.to_s, err.backtrace
     end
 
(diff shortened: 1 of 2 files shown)


bzwei:
@miq-bot assign @gmcculloug 
@miq-bot add_label bug
@miq-bot add_label gaprindashvili/yes

bzwei:
@mkanoor please review

app/models/manageiq/providers/ansible_tower/shared/automation_manager/job.rb
```diff
@@ -31,6 +32,23 @@ def db_name
     def status_class
       "#{self.name}::Status".constantize
     end
+
+    private
+
+    def filter_options(template, options)
```
mkanoor: @bzwei 
Can we name this function to indicate its purpose.
e.g. transform_extra_vars_keys or convert_extra_vars_keys

mkanoor:
@bzwei 
Can we add a single line comment there to indicate that Automate is case insensitive and we loose the actual name of extra_var being used in the playbook and this function rectifies that.

mkanoor:
@bzwei How about using reconcile_extra_var_keys?

simaishi:
Gaprindashvili backport details:

```
$ git log -1
commit 0e04700ac16df2d36f25774d90b778c4bc76ec6f
Author: Greg McCullough <gmccullo@redhat.com>
Date:   Wed Nov 8 10:11:08 2017 -0500

    Merge pull request #34 from bzwei/extra_var_case
    
    Correct extra_var keys to original case
    (cherry picked from commit e264825a855563b894e49a419f5255f36e356a48)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1511976
```

PR #41, opened 2017-12-07
[WIP] orchestrate_destroy manager before_destroy

jameswnl:
https://bugzilla.redhat.com/show_bug.cgi?id=1491704

Destroying Ansible Tower Provider is being rolled back because the associated manager is being held up by workers (introduced by #14675)


jameswnl:
Being done in https://github.com/ManageIQ/manageiq/pull/16614

PR #46, opened 2017-12-13
Allow Satellite6Credential to be removed from Embedded Ansible space

jameswnl:
https://bugzilla.redhat.com/show_bug.cgi?id=1525172

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/shared/inventory/parser/automation_manager.rb
+++ b/app/models/manageiq/providers/ansible_tower/shared/inventory/parser/automation_manager.rb
@@ -74,27 +74,33 @@ module ManageIQ::Providers::AnsibleTower::Shared::Inventory::Parser::AutomationM
     end
   end
 
+  def miq_credential_types
+    @miq_credential_types ||= begin
+      provider_module = ManageIQ::Providers::Inflector.provider_module(collector.manager.class).name
+      supported_types = "#{provider_module}::AutomationManager::Credential".constantize.descendants.collect(&:name)
+      {
+        'net'        => "#{provider_module}::AutomationManager::NetworkCredential",
+        'ssh'        => "#{provider_module}::AutomationManager::MachineCredential",
+        'vmware'     => "#{provider_module}::AutomationManager::VmwareCredential",
+        'scm'        => "#{provider_module}::AutomationManager::ScmCredential",
+        'aws'        => "#{provider_module}::AutomationManager::AmazonCredential",
+        'rax'        => "#{provider_module}::AutomationManager::RackspaceCredential",
+        'satellite6' => "#{provider_module}::AutomationManager::Satellite6Credential",
+        'gce'        => "#{provider_module}::AutomationManager::GoogleCredential",
+        'azure'      => "#{provider_module}::AutomationManager::AzureClassicCredential",
+        'azure_rm'   => "#{provider_module}::AutomationManager::AzureCredential",
+        'openstack'  => "#{provider_module}::AutomationManager::OpenstackCredential"
+      }.select { |_tower_type, miq_type| supported_types.include?(miq_type) }
+    end
+  end
+
   def credentials
+    provider_module = ManageIQ::Providers::Inflector.provider_module(collector.manager.class).name
     collector.credentials.each do |credential|
       inventory_object = persister.credentials.find_or_build(credential.id.to_s)
       inventory_object.name = credential.name
       inventory_object.userid = credential.username
-      provider_module = ManageIQ::Providers::Inflector.provider_module(collector.manager.class).name
-      inventory_object.type = case credential.kind
-                                when 'net' then "#{provider_module}::AutomationManager::NetworkCredential"
-                                when 'ssh' then "#{provider_module}::AutomationManager::MachineCredential"
-                                when 'vmware' then "#{provider_module}::AutomationManager::VmwareCredential"
-                                when 'scm' then "#{provider_module}::AutomationManager::ScmCredential"
-                                when 'aws' then "#{provider_module}::AutomationManager::AmazonCredential"
-                                when 'rax' then "#{provider_module}::AutomationManager::RackspaceCredential"
-                                when 'satellite6' then "#{provider_module}::AutomationManager::Satellite6Credential"
-                                # when 'cloudforms' then "#{provider_module}::AutomationManager::$$$Credential"
-                                when 'gce' then "#{provider_module}::AutomationManager::GoogleCredential"
-                                when 'azure' then "#{provider_module}::AutomationManager::AzureClassicCredential"
-                                when 'azure_rm' then "#{provider_module}::AutomationManager::AzureCredential"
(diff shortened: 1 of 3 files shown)


jameswnl:
@miq-bot add_labels bug, blocker

spec/models/manageiq/providers/ansible_tower/automation_manager/refresher_spec.rb
```diff
@@ -1,4 +1,10 @@
 describe ManageIQ::Providers::AnsibleTower::AutomationManager::Refresher do
+  let(:more_credential_types) do
+    {
+      'hello_sat_cred' => 'Satellite6Credential'
```
blomquisg: Is this expected to be a setting that is configured in `settings.yml`?

If so, add a comment to the settings file in this repo to show that what setting would look like if someone wanted to override it.

jameswnl: Greg, this is not a setting... and hmm... I'm not sure you ask this way...

This refresher test is based on a cassette which contains of all types of credential types. This used to be in for both embedded and external one. I'm pulling this out here so that embedded won't fail.

app/models/manageiq/providers/ansible_tower/shared/inventory/parser/automation_manager.rb
```diff
@@ -74,27 +74,33 @@ def configuration_script_sources
     end
   end
 
+  def miq_credential_type(tower_type)
```
blomquisg: Recommend `tower_credential_type` to be consistent with the method name.

jameswnl: thanks! Done

app/models/manageiq/providers/ansible_tower/shared/inventory/parser/automation_manager.rb
```diff
@@ -74,27 +74,33 @@ def configuration_script_sources
...
+    @provider_module ||= ManageIQ::Providers::Inflector.provider_module(collector.manager.class).name
+    @miq_credential_type ||= begin
```
blomquisg: I was confused by this ivar name matching the method name because the ivar is an array, but the method returns a single value.

Here's what I suggest doing.

1. Create a separate ivar-bound method called `miq_credential_types` that moves all the code form lines 78-94 in this method (in that method, move `provider_module` inside the `begin` block then you don't have to have another ivar.
2. Then, this method becomes:

```ruby
def miq_credential_type(tower_credential_type)
  miq_credential_types[tower_credential_type] || "#{@provider_module}::AutomationManager::Credential"
end
```

That way, you have one method that just loads the array `miq_credentials_types`, which is bound to an ivar called `@miq_credential_types`.  Then, you have another method that uses that cached list of credential types to look up the appropriate credential, or returns the default.

app/models/manageiq/providers/ansible_tower/shared/inventory/parser/automation_manager.rb
```diff
@@ -74,27 +74,33 @@ def configuration_script_sources
     end
   end
 
+  def tower_credential_type(tower_type)
+    @provider_module ||= ManageIQ::Providers::Inflector.provider_module(collector.manager.class).name
+    @tower_credential_type ||= begin
```
blomquisg: D'oh, you changed the wrong thing ... I guess I wasn't clear.

I was thinking that `miq_credential_type` was fine for the method name.  But, the *param* name should change to `tower_credential_type` for symmetry.

However, read on...

I was confused by this ivar name matching the method name because the ivar is an array, but the method returns a single value.

Here's what I suggest doing.

1. Create a separate ivar-bound method called `miq_credential_types` that moves all the code form lines 78-94 in this method (in that method, move `provider_module` inside the `begin` block then you don't have to have another ivar.
2. Then, this method becomes:

```ruby
def miq_credential_type(tower_credential_type)
  miq_credential_types[tower_credential_type] || "#{@provider_module}::AutomationManager::Credential"
end
```

That way, you have one method that just loads the array `miq_credentials_types`, which is bound to an ivar called `@miq_credential_types`.  Then, you have another method that uses that cached list of credential types to look up the appropriate credential, or returns the default.

spec/models/manageiq/providers/ansible_tower/automation_manager/refresher_spec.rb
blomquisg: K, I was actually confused by `hello_sat_cred`.  I thought that was some dummy name.  But I found it in the actual code for defining the Sat6 Credential.

I understand how this is being used now.

app/models/manageiq/providers/ansible_tower/shared/inventory/parser/automation_manager.rb
jameswnl: it actually began like that. and then i don't want to keep looking up this `provider_module` again. see my next version...

app/models/manageiq/providers/ansible_tower/shared/inventory/parser/automation_manager.rb
```diff
@@ -74,27 +74,32 @@ def configuration_script_sources
     end
   end
 
+  def miq_credential_types(provider_module)
```
blomquisg: This isn't good.

It says that the first time you call this method with a provider_module, it will cache the `miq_credential_types` with that provider module.  If you call again with a different provider module, it will give you the old value.

Refactoring this out is probably creating too much of a problem.  You're going to keep digging down until you have a solution that's more complicated than the problem it's trying to solve.

You should probably just inline this again in the `credentials` method, and don't cache the result in `miq_credential_types`.

If we find that's causing some performance issues one day, we can revisit this.

jameswnl: This parser is an instance and shouldn't be used for other `provider_module`

app/models/manageiq/providers/ansible_tower/shared/inventory/parser/automation_manager.rb
```diff
@@ -74,27 +74,33 @@ def configuration_script_sources
     end
   end
 
+  def miq_credential_types
+    @miq_credential_types ||= begin
+      provider_module = ManageIQ::Providers::Inflector.provider_module(collector.manager.class).name
+      supported_types = "#{provider_module}::AutomationManager::Credential".constantize.descendants.collect(&:name)
```
agrare: That's slick

simaishi:
Gaprindashvili backport details:

```
$ git log -1
commit bb92de8247db3160a597070dca267944b2ff7bd3
Author: Adam Grare <agrare@redhat.com>
Date:   Tue Dec 19 08:33:46 2017 -0500

    Merge pull request #46 from jameswnl/sat6
    
    Allow Satellite6Credential to be removed from Embedded Ansible space
    (cherry picked from commit d6ba79104f93971ecf5b50bcfc9e18d8f5ec9a5b)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1527583
```

PR #52, opened 2018-01-10
WIP: testing dependent-destroy

jameswnl:
See the different behaviors of the 2 added specs.

### DIFF
--- a/spec/support/ansible_shared/provider.rb
+++ b/spec/support/ansible_shared/provider.rb
@@ -36,6 +36,46 @@ shared_examples_for "ansible provider" do
       expect(OperatingSystem.count).to       eq(0)
       expect(Hardware.count).to              eq(0)
     end
+
+    it "thru automation_manager will remove all child objects" do
+      subject.automation_manager.configured_systems = [
+        FactoryGirl.create(:configured_system, :computer_system =>
+          FactoryGirl.create(:computer_system,
+                             :operating_system => FactoryGirl.create(:operating_system),
+                             :hardware         => FactoryGirl.create(:hardware)))
+      ]
+
+      expect(ExtManagementSystem.count).to eq(1)
+      expect(subject.automation_manager.class.name).to eq("ManageIQ::Providers::AnsibleTower::AutomationManager")
+      subject.automation_manager.destroy
+
+      expect(Provider.count).to              eq(1)
+      expect(ExtManagementSystem.count).to   eq(0)
+      expect(ConfiguredSystem.count).to      eq(0)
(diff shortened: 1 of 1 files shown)


Fryguy:
I just don't get it...need to investigate more :/

Ladas:
the data are not stored properly

check 
```
subject.automation_manager.reload.configured_systems
```

D, [2018-01-10T21:22:01.373714 #12596] DEBUG -- :   ManageIQ::Providers::AutomationManager::ConfiguredSystem Load (0.3ms)  SELECT "configured_systems".* FROM "configured_systems" WHERE "configured_systems"."type" IN ('ManageIQ::Providers::AutomationManager::ConfiguredSystem', 'ManageIQ::Providers::ExternalAutomationManager::ConfiguredSystem', 'ManageIQ::Providers::EmbeddedAutomationManager::ConfiguredSystem', 'ManageIQ::Providers::AnsibleTower::AutomationManager::ConfiguredSystem', 'ManageIQ::Providers::EmbeddedAnsible::AutomationManager::ConfiguredSystem') AND "configured_systems"."manager_id" = $1  [["manager_id", 2952]]

The configured system needs to have valid STI type, otherwise the query will not find it. subject.automation_manager.configured_systems is not doing any query, it just shows cached data

jameswnl:
@Ladas see my fix here https://github.com/ManageIQ/manageiq/pull/16798

spec/support/ansible_shared/provider.rb
```diff
@@ -36,6 +36,46 @@
       expect(OperatingSystem.count).to       eq(0)
       expect(Hardware.count).to              eq(0)
     end
+
+    it "thru automation_manager will remove all child objects" do
+      subject.automation_manager.configured_systems = [
+        FactoryGirl.create(:configured_system, :computer_system =>
```
Ladas: you should use different factory https://github.com/Ladas/manageiq/blob/2f0ef1a90758f2b5c37e93a0ed2f586a916dad64/spec/factories/configured_system.rb#L10

or the default factory should contain valid base class STI, in this case ManageIQ::Providers::AutomationManager::ConfiguredSystem

-----

None of factories should be producing a record that can't be created in real life, e.g. ConfiguredSystem with type nil is not valid, refresh always set :type

jameswnl:
thanks to @chrisarcand problem solved, closing this

spec/support/ansible_shared/provider.rb
chrisarcand: Thanks @Ladas !

PR #60, opened 2018-02-20
Remove dependency on custom have_attributes

chrisarcand:
In core we have a custom `have_attributes` matcher that overrides the
normal RSpec one. Among other silly things, it allows you to pass a time
object and assume that you want to approximate it (which was yet another
custom matcher, too...).

We are removing these matchers to reduce confusion and provide the
actual RSpec ones which are more feature rich anyway.

Things like time approximation can and should be done in the caller
doing something like this (see code).

Needs https://github.com/ManageIQ/manageiq/pull/17021


chrisarcand:
@miq-bot assign @bdunne

chrisarcand:
This needs to be merged at the same time as https://github.com/ManageIQ/manageiq/pull/17021 ; you can't make this build green without it. Funny enough, you can't make the other one green without this one  🙃

PR #72, opened 2018-04-12
Get stdout on ansible repository refresh

Glutexo:
If a failed Ansible repository is found during Tower refresh, fetch and store stdout capture of its last update job.

Fetching and presenting this information is a workaround for cloning repositories from HTTPS servers with untrusted SSL certificates. Tower API does not offer an option to disable SSL check, so with the stdout user can at least see, why the cloning actually failed.

To better communicate what’s going on when adding a new repository, split add success/failure notification in two parts:
* Project has been successfully created in provider
* Project update job (first clone) succeeded

With legacy Tower versions (like 3.0.2), the stdout capture is almost never available. This is because we trigger an unnecessary update right after creating the project in the provider, when Tower is not ready to store the capture yet. This will be fixed in a further PR.

https://bugzilla.redhat.com/show_bug.cgi?id=1513616

Prior merging this PR, following prerequisites have to be met:

* [x] ansible/ansible_tower_client_ruby#102 – merged
* [x] ansible/ansible_tower_client_ruby v0.14.0 – released
* [x] manageiq/manageiq#17290

HTTP requests issued by tests changed. Because of that VCR cassettes had to be re-recorded. A rake task that populates the data used by the tests needs improvements. These are subject of another PR #73.

/cc @jameswnl for review

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/shared/inventory/parser/automation_manager.rb
+++ b/app/models/manageiq/providers/ansible_tower/shared/inventory/parser/automation_manager.rb
@@ -1,5 +1,5 @@
 module ManageIQ::Providers::AnsibleTower::Shared::Inventory::Parser::AutomationManager
-  ERROR_MAX_SIZE = 50 * 1024 # 50 KiB
+  ERROR_MAX_SIZE = 50.kilobytes
 
   def parse
     inventory_root_groups
@@ -67,13 +67,11 @@ module ManageIQ::Providers::AnsibleTower::Shared::Inventory::Parser::AutomationM
       inventory_object.scm_update_on_launch = project.scm_update_on_launch
       inventory_object.status = project.status
       inventory_object.last_updated_on = project.last_updated
-
       inventory_object.last_update_error = nil
-      unless project.status == 'successful'
+
+      unless inventory_object.status == 'successful'
         last_update = project.last_update
-        if last_update
-          inventory_object.last_update_error = last_update.result_stdout[0..ERROR_MAX_SIZE]
-        end
+        inventory_object.last_update_error = last_update.result_stdout.mb_chars.limit(ERROR_MAX_SIZE) if last_update
       end
 
       project.playbooks.each do |playbook_name|
--- a/app/models/manageiq/providers/ansible_tower/shared/inventory/parser/automation_manager.rb
+++ b/app/models/manageiq/providers/ansible_tower/shared/inventory/parser/automation_manager.rb
@@ -1,4 +1,6 @@
 module ManageIQ::Providers::AnsibleTower::Shared::Inventory::Parser::AutomationManager
+  ERROR_MAX_SIZE = 50 * 1024 # 50 KiB
+
   def parse
     inventory_root_groups
     configured_systems
@@ -65,6 +67,14 @@ module ManageIQ::Providers::AnsibleTower::Shared::Inventory::Parser::AutomationM
       inventory_object.scm_update_on_launch = project.scm_update_on_launch
       inventory_object.status = project.status
       inventory_object.last_updated_on = project.last_updated
+
+      inventory_object.last_update_error = nil
+      unless project.status == 'successful'
+        last_update = project.last_update
+        if last_update
+          inventory_object.last_update_error = last_update.result_stdout[0..ERROR_MAX_SIZE]
+        end
+      end
 
       project.playbooks.each do |playbook_name|
         inventory_object_playbook = persister.configuration_script_payloads.find_or_build_by(
--- a/spec/support/ansible_shared/automation_manager/refresh_configuartion_script_source.rb
+++ b/spec/support/ansible_shared/automation_manager/refresh_configuartion_script_source.rb
@@ -20,10 +20,10 @@ shared_examples_for "refresh configuration_script_source" do |ansible_provider, 
     configuration_script_source = FactoryGirl.create(:"#{ems_type}_configuration_script_source",
                                                      :authentication => credential,
                                                      :manager        => automation_manager,
-                                                     :manager_ref    => 325)
+                                                     :manager_ref    => 472)
     configuration_script_source.configuration_script_payloads.create!(:manager_ref => '2b_rm', :name => '2b_rm')
     configuration_script_source_other = FactoryGirl.create(:"#{ems_type}_configuration_script_source",
-                                                           :manager_ref => 322,
+                                                           :manager_ref => 5,
                                                            :manager     => automation_manager,
                                                            :name        => 'Dont touch this')
 
@@ -31,7 +31,7 @@ shared_examples_for "refresh configuration_script_source" do |ansible_provider, 
     stub_const("ManageIQ::Providers::AnsibleTower::Shared::AutomationManager::ConfigurationScriptSource::REFRESH_ON_TOWER_SLEEP", 0.seconds)
 
     # this is to check if a project will be updated on tower
-    last_project_update = Time.zone.parse("2018-05-23 09:29:41.7849700 +0000") - 1.minute
+    last_project_update = Time.zone.parse("2017-04-10 20:50:11.429285000 +0000") - 1.minute
 
     Spec::Support::VcrHelper.with_cassette_library_dir(ManageIQ::Providers::AnsibleTower::Engine.root.join("spec/vcr_cassettes")) do
       2.times do
@@ -48,23 +48,23 @@ shared_examples_for "refresh configuration_script_source" do |ansible_provider, 
           expect(last_updated).to be >= last_project_update
           last_project_update = last_updated
 
-          expect(configuration_script_source.name).to eq("hello_repo")
+          expect(configuration_script_source.name).to eq("targeted_refresh")
           expect(configuration_script_source.last_updated_on).to eq(last_updated)
-          expect(ConfigurationScriptPayload.count).to eq(61)
+          expect(ConfigurationScriptPayload.count).to eq(81)
           expect(ConfigurationScriptPayload.where(:name => '2b_rm')).to be_empty
-          expect(configuration_script_source.configuration_script_payloads.count).to eq(61)
+          expect(configuration_script_source.configuration_script_payloads.count).to eq(81)
           expect(
             configuration_script_source.configuration_script_payloads.where(
-              :name => "jboss-standalone/demo-aws-launch.yml"
+              :name => "test/utils/docker/httptester/httptester.yml"
             ).count
           ).to eq(1)
-          expect(configuration_script_source.authentication.name).to eq('hello_scm_cred')
+          expect(configuration_script_source.authentication.name).to eq('db-github')
           expect(credential.reload).to eq(credential)
 
           expect(configuration_script_source_other.name).to eq("Dont touch this")
         end
         # check if a playbook will be added back in on the second run
-        configuration_script_source.configuration_script_payloads.where(:name => "jboss-standalone/demo-aws-launch.yml").destroy_all
+        configuration_script_source.configuration_script_payloads.where(:name => "test/utils/docker/httptester/httptester.yml").destroy_all
       end
     end
   end
--- a/spec/support/ansible_shared/automation_manager/refresh_configuartion_script_source.rb
+++ b/spec/support/ansible_shared/automation_manager/refresh_configuartion_script_source.rb
@@ -20,10 +20,10 @@ shared_examples_for "refresh configuration_script_source" do |ansible_provider, 
     configuration_script_source = FactoryGirl.create(:"#{ems_type}_configuration_script_source",
                                                      :authentication => credential,
                                                      :manager        => automation_manager,
-                                                     :manager_ref    => 472)
+                                                     :manager_ref    => 325)
     configuration_script_source.configuration_script_payloads.create!(:manager_ref => '2b_rm', :name => '2b_rm')
     configuration_script_source_other = FactoryGirl.create(:"#{ems_type}_configuration_script_source",
-                                                           :manager_ref => 5,
+                                                           :manager_ref => 322,
                                                            :manager     => automation_manager,
                                                            :name        => 'Dont touch this')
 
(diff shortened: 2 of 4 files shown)


app/models/manageiq/providers/ansible_tower/shared/automation_manager/tower_api.rb
```diff
@@ -76,9 +83,12 @@ def update_in_provider(params)
     with_provider_object do |provider_object|
       provider_object.update_attributes!(params)
     end
+
     if respond_to?(:refresh_in_provider)
-      refresh_in_provider
+      refresh_success = refresh_in_provider
+      self.class.send('notify_refresh_in_provider', manager_id, params, refresh_success)
```
jrafanie: Why not make this public so we don't have to use `send`?

jrafanie:
Note, @Glutexo this PR should include the version of the ansible tower client ruby gem that includes this [PR](https://github.com/ansible/ansible_tower_client_ruby/pull/102)

Somehow, this repo doesn't rely on the client in the `gemspec`?

https://github.com/ManageIQ/manageiq-providers-ansible_tower/blob/e2a7e55f13330cd4cc8e4430ef541bba843d2650/manageiq-providers-ansible_tower.gemspec#L15

I'm confused.  But either way, if this repo relies on gem version, we should list in the gemspec and bump to the new version that includes the change needed.

app/models/manageiq/providers/ansible_tower/shared/automation_manager/tower_api.rb
jrafanie: Ok, so why are this new method and `notify` both private class methods that we want to call from an instance?  If they were all instance methods, you can call them directly without `send` or `self` and the "internal consumption" purpose can be maintained.  If you want to call them from both class scope and instance scope, they're not private methods.

jrafanie: Note, I'm not saying to fix this all right now in this PR but the usage of `send` indicates improper visibility of this method.

Glutexo:
@jrafanie Even though this repository has a gemspec, it is [not used as a normal gem](https://github.com/ManageIQ/manageiq/blob/9f8c439882b88b6dc0866d74c0df29c168cdaca7/Gemfile#L20). It’s a ManageIQ plugin and the dependency is stated in the [manageiq/manageiq Gemfile](https://github.com/ManageIQ/manageiq/blob/9f8c439882b88b6dc0866d74c0df29c168cdaca7/Gemfile#L89) Thanks for reminding me to bump the version there!

Fryguy:
@Glutexo The reason that the client dependency is in the main ManageIQ repo is because that dependency is for the embedded_ansible provider, which is still in the main repo.  This repo *must* have the client dependency listed in the gemspec, since this repo depends on it directly.  The fact that it's loaded via the [ManageIQ Gemfile](https://github.com/ManageIQ/manageiq/blob/9f8c439882b88b6dc0866d74c0df29c168cdaca7/Gemfile#L89) doesn't really matter because that reference will still point to this repo's gemspec.

app/models/manageiq/providers/ansible_tower/shared/inventory/parser/automation_manager.rb
```diff
@@ -65,6 +65,10 @@ def configuration_script_sources
       inventory_object.status = project.status
       inventory_object.last_updated_on = project.last_updated
 
+      if project.status == 'failed'
```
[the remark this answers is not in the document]
Glutexo: Good point! I’ll update `last_update_stdout` every time. Since the output capture is presented on a separate subpage, it does not disturb even for the success. Still, it can contain some useful information like warnings and such.

app/models/manageiq/providers/ansible_tower/shared/automation_manager/tower_api.rb
Glutexo: The method visibility is not all that wrong. It is accessed only from the `TowerApi` class itself or its instance. It must be called via `send` though, because it is defined as private inside the `ClassMethods` module.

This is however not my invention: I used the style already used in the class. The `notify`, `queue` etc. methods are in that space too and so are they called using `send`.

app/models/manageiq/providers/ansible_tower/shared/automation_manager/tower_api.rb
```diff
@@ -44,6 +47,10 @@ def notify(op, manager_id, params, success)
       )
     end
 
+    def notify_refresh_in_provider(manager_id, params, success)
+      notify('refresh in provider', manager_id, params, success)
```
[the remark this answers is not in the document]
Glutexo: It would definitely work like that. However I decided to make it a method, so the `'refresh in provider'` string describing the notification is only in one place. It’s like the `notify('creation', …)`, but this `'refresh in provider'` is called from two places: after create and after update.

Glutexo:
I‘ve updated the PR. The `last_update_stdout` capture is now saved every time a Project refreshes. I kept the method visibility and method extraction as they were, because I believe they are justified – see my replies to the comments. If you think it’s really not good, I can make more changes though.

There is now a `add_runtime_dependency` in the `Gemspec` requiring `ansible_tower_client, "~ 0.14.0"`. This version has not been released yet, we must wait for the release.
It is possible that this gem has more runtime dependencies. I however added just this one as it is related to this PR’s changes.

Rebased on #73, which (if @jameswnl doesn’t see any problems with that) should be ready to merge by now.

Glutexo:
[ansible/ansible_tower_client_ruby v0.14.0](https://github.com/ansible/ansible_tower_client_ruby/tree/v0.14.0) needed for this PR has been released.

Glutexo:
Rebased on current master and currrent #73.

app/models/manageiq/providers/ansible_tower/shared/automation_manager/tower_api.rb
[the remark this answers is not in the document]
Glutexo: Okay, I’ll remove the method and call `notify` directly. I just liked the method more than calling `notify` with the same string from two different places. To me it conveyed a message saying “this is the same thing” making clear it is an intention. The string argument on the other hand is arbitrary and could be the same only by accident.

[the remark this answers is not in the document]
Glutexo: Okay, I’ll remove the method and call `notify` directly. I just liked the method more than calling `notify` with the same string from two different places. To me it conveyed a message saying “this is the same thing” making clear it is an intention. The string argument on the other hand is arbitrary and could be the same only by accident.

Glutexo:
If #84 would get merged before this PR, it is necessary to bump the Ansible Tower Client version in the Gemspec to `v0.14.0` like this:

```
s.add_runtime_dependency "ansible_tower_client", "~> 0.43.0"
```

Glutexo:
@jameswnl This PR requires VCR re-recording. Because of that it is based on #73, which fixed the Population Tower Rake task and makes tests use solely the populated entities. Please get that one merged first, then we can jump to this one.

Glutexo:
As @jameswnl suggested, I untangled this PR from dependency on #73. The helper Rake task changes are not subject of the issue itself. VCRs still have to be re-recorded as getting the stdouts issues new requests for every project upon refresh.

Glutexo:
I have found out that this would fail with Tower v2 in case there is no last update for a given project. [`project.last_update`][1] raises an error in such case. On the other hand with Tower v3 [this][1] just returns `nil`, so everything works fine.

As @bascard pointed out in his [BZ comment][2], Tower v2 has reached its [End of Life][3] a long time ago and thus there’s no need to fix this. Still it would be good to mention this in the docs or errata docs. Unfortunately I am not sure where to put such a note. May I ask for some directions where to find the appropriate docs? (@jameswnl?)

[1]: https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/72/files#diff-837df86d6305c7efe0df21cabde606e0R67
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1513616#c40
[3]: https://access.redhat.com/support/policy/updates/ansible-tower

Glutexo:
@jameswnl confirmed that there is no need to care about the legacy version of Tower. Dropping its support [has already started](https://github.com/ManageIQ/manageiq/issues/17491). That said, we are waiting only for https://github.com/ManageIQ/manageiq/pull/17290 to get merged.

jameswnl:
@Glutexo I think pretty much there once you rebase and add the `s.add_runtime_dependency "ansible_tower_client", "~> 0.14.0"`

Glutexo:
#84 is now merged, so rebased and bumped the required `ansible_tower_client_ruby` gem version to `~> v0.14.0`.

Glutexo:
Rebased, resolved conflict and re-recorded a VCR that is needed for testing the stdout retrieval.

Incorporated changes requested by @Fryguy: https://github.com/ManageIQ/manageiq/pull/17290#issuecomment-398781641

* Renamed the field to `last_update_error`.
* The stdout is saved only on failure.
* The stdout is capped on an arbitrary length of 50 KiB, just in case.

This might require a new review by @jameswnl and should reduce work left for @tumido.

app/models/manageiq/providers/ansible_tower/shared/inventory/parser/automation_manager.rb
```diff
@@ -66,6 +68,14 @@ def configuration_script_sources
       inventory_object.status = project.status
       inventory_object.last_updated_on = project.last_updated
 
+      inventory_object.last_update_error = nil
+      unless project.status == 'successful'
+        last_update = project.last_update
```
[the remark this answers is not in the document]
Glutexo: `project`.`last_update` triggers a resource download from the API. As far as I know the result is not cached. Using `project.last_update` twice would fire two requests.

app/models/manageiq/providers/ansible_tower/shared/inventory/parser/automation_manager.rb
```diff
@@ -66,6 +68,14 @@ def configuration_script_sources
...
+        if last_update
+          inventory_object.last_update_error = last_update.result_stdout[0..ERROR_MAX_SIZE]
```
tumido: @Glutexo do you know if the stdout string comes as ASCII or can it be multibyte characters?

I'd prefer this to be changed to something like:
```ruby
inventory_object.last_update_error = last_update.result_stdout.mb_chars.limit(ERROR_MAX_SIZE)
```
(Maybe `to_s` would be needed at the end. I haven't tried to save it into DB yet.)

Glutexo: Since it is a console output, it can definitely contain UTF-8 characters and I’d expect it to come as UTF-8 text. If it doesn’t, I’d consider it a bug. You can at least check the Ruby encoding of the string right before the save: `p(last_update.result_stdout.encoding)`. If it’s really UTF-8, then we can safely use `[0..n]`. Otherwise we’d have to investigate further what the real encoding is.

tumido: I agree. In that case we shouldn't be using slices, since it counts letters, not bytes. 50kb != 50letters. The `limit` method on `mb_chars` is encoding safe, and limits the length to certain amount fitting the last character in (no matter how many bytes it is for single letter).

Glutexo: You’re right. I was allured by the simplicity of `[]` so much, that I forgot that byte to char count does not map 1:1. (That me, who gives talks at conferences about character encodings and internationalization. What a shame! ☹️) Thanks for letting me know about [`limit`](http://api.rubyonrails.org/v5.0.6/)!

jameswnl:
LGTM
Waiting for the 2 dependent PRs

app/models/manageiq/providers/ansible_tower/shared/inventory/parser/automation_manager.rb
```diff
@@ -65,6 +67,12 @@ def configuration_script_sources
       inventory_object.scm_update_on_launch = project.scm_update_on_launch
       inventory_object.status = project.status
       inventory_object.last_updated_on = project.last_updated
+      inventory_object.last_update_error = nil
+
+      unless inventory_object.status == 'successful'
+        last_update = project.last_update
+        inventory_object.last_update_error = last_update.result_stdout.mb_chars.limit(ERROR_MAX_SIZE) if last_update
```
Fryguy: Curious on why you chose `.mb_chars.limit` over just `.truncate`.  We use the latter, so it's preferable unless you have a specific reason.

Fryguy: If you're trying to solve issues with result_stdout being non-UTF-8 (US-ASCII or ASCII-8BIT) and being afraid of chopping multibyte chars in half, you could also just do the following

```ruby
result_stdout.force_encoding("UTF-8").truncate(ERROR_MAX_SIZE)
```

(mb_chars assumes your incoming String is always UTF-8 anyway)

Fryguy: Additionally, I think you need a spec for this truncation.  If you `stub_const(ManageIQ::Providers::AnsibleTower::Shared::Inventory::Parser::AutomationManager::ERROR_MAX_SIZE, 10)` or some really small number, then you can see if it truncates it to that amount.

Fryguy: cc @tumido

tumido: @Fryguy I've chose to use `.mb_chars.limit` over `.truncate` because we want to limit the data to a certain size not to a certain string length. [`.truncate` counts characters not bytes](https://apidock.com/rails/String/truncate). If we want to use `truncate` we shouldn't have a constant named `ERROR_MAX_SIZE` but rather `ERROR_MAX_LENGTH` and it shouldn't be in bytes...

```ruby
> "ááááááááááá".mb_chars.limit(8).bytesize
8
> "ááááááááááá".truncate(8).bytesize
13
```

tumido:
@Fryguy, I've added the spec finally! Also - the explanation about the truncation I've provided above - I still think this is a valid approach, but if you would insist and rather see it being changed to `truncate`, I can do so. Just let me know. :wink:

tumido:
@Glutexo any change you have an updated `refresher_targeted_configuration_script_source.yml` VCR cassette laying around? It seems you've forgotten to include it into this PR.

Glutexo:
@tumido It looks like it’s not the VCR file that is missing, but actually the spec/support/ansible_shared/automation_manager/refresh_configuartion_script_source.rb that is redundant. It contains only update so variable Tower data, but not changes in logic. I have probably left that in the PR only by mistake. Feel free to revert it.

tumido:
@Fryguy , @jrafanie , @Ladas , @jameswnl I think this one is ready to be reviewed once more. I think all comments and concerns have been addressed now.

PR #73, opened 2018-04-17
Update Populate Tower Rake task

Glutexo:
Update specs and a Rake task that populates data for them.

* Do not use removed credential kinds: Azure (Classic) and Rackspace.
* Wait until the created project finishes cloning its repository. Only after this succeeds, move on to creation of job templates that depend on the project.
* Do not depend on any objects in a Tower instance that are not populated by the Rake task. Add instructions to the task for those that can’t be created via the API.
* Re-record VCR cassettes for the updated specs.

This allows to re-record VCR cassettes any time, when the specs change. (Currently required by #72) That was not possible because of the non-documented objects that were expected in the Tower instance. Removing the credentials also allows to use more recent Tower 3.2.2 for the recording.

/cc @jameswnl for review.


lib/tasks_private/spec_helper.rake
```diff
@@ -173,20 +165,27 @@ class PopulateTower
...
+      # create a job_template with survey spec
+      uri = '/api/v1/job_templates/'
+      data = {"name" => "hello_template_with_survey", "description" => "test job with survey spec", "job_type" => "run", "project" => project['id'], "playbook" => "hello_world.yml", "credential" => machine_credential['id'], "inventory" => inventory['id'], "survey_enabled" => true, "organization" => organization['id']}
+      template = create_obj(uri, data)
+      # create survey spec
+      uri = "/api/v1/job_templates/#{template['id']}/survey_spec/"
+      data = {"name" => "Simple Survey", "description" => "Description of the simple survey", "spec" => [{"type" => "text", "question_name" => "example question", "question_description" => "What is your favorite color?", "variable" => "favorite_color", "required" => false, "default" => "blue"}]}
+      @conn.post(uri, data).body
+      puts "created #{template['url']} survey_spec"
+    rescue Faraday::ClientError => error
```
jameswnl: First call's failure will now short-circuit the other 2 creations.  If you want to improve this, probably better to
 * put a sleep before this section
 * and/or enclose each individual creation with `rescue` and do a retry for each

lib/tasks_private/spec_helper.rake
```diff
@@ -173,20 +165,27 @@ class PopulateTower
...
+      puts "Project 'hello_repo' probably hasn’t finished syncing yet." if error.response[:status] == 400
```
jameswnl: You could be catching creation of `hello_template` or `hello_template_with_survey`, so the name needs to be a variable.

lib/tasks_private/spec_helper.rake
```diff
@@ -142,18 +142,10 @@ class PopulateTower
...
     data = {"name" => "hello_azure_cred", "kind" => "azure_rm", "username" => "admin", "password" => "abc", "subscription"  => "sub_id", "tenant" => "ten_id", "secret" => "my_secret", "client" => "cli_id", "organization" => organization['id']}
     _azure_credential = create_obj(uri, data)
 
-    # create cloud azure(Classic) cred
-    data = {"name" => "hello_azure_classic_cred", "kind" => "azure", "username" => "admin", "ssh_key_data" => ssh_key_data, "organization" => organization['id']}
-    _azure_credential = create_obj(uri, data)
-
     # create cloud satellite6 cred
     data = {"name" => "hello_sat_cred", "kind" => "satellite6", "username" => "admin", "password" => "abc", "host"  => "s1.sat.com", "organization" => organization['id']}
     _azure_credential = create_obj(uri, data)
```
jameswnl: A favor to ask for a knit that's not related to your changes: can you rename this `_azure_credential` to `_sat6_credential`?

Glutexo:
There are more problems with this. The VCR’s are practically unre-recordable in their current state: 

The general problem is that some examples expect some objects to be present in Tower, that are not created by the Rake task. I’ve already fixed those that were fine with using those already created by the Rake task. Those that require more objects to be created still wait to be fixed.

I’m going to put those fixes into this PR along with the changes based on your comments.

lib/tasks_private/spec_helper.rake
Glutexo: Sure. Updated that, I will include it with rebase.

lib/tasks_private/spec_helper.rake
Glutexo: You’re right and this was my original thought. I somehow got the impression that you’re not a fan of the retry-wait flow for such a one-purpose tool. It was actually not that complex, so I added it instead of this short-circuit. I’ll push it right away.

Glutexo:
@jameswnl 

I made the changes you requested:
* Replaced the short-circuit failing to wait-(re)try flow.
* Renamed the sat6 variable.

Then I updated the specs so they don’t depend on objects not created by the Rake task. The only exception is a Vault credential that can’t be created via the API: added instructions to the Rake task to create it manually after populating.

With the updated specs I re-recorded the VCR cassettes using the test Tower 3.2.2 instance. @miq-bot complaints about some credentials that have “cloudforms” in their host URL’s. Is this a problem? If so, I can remove them from the shared Tower instance and make one more re-recording. I just want to be sure that nobody uses them for anything.

Is the PR ok for you now?

Glutexo:
One more force push: I actually replaced the private hostnames with `example.com` in the cassettes. They are not used in the tests, since they are only part of the unrelated data sitting there in the Tower instance.

lib/tasks_private/spec_helper.rake
```diff
@@ -204,6 +243,24 @@ class PopulateTower
...
+  end
+
+  def information
+    return unless @organization
+
+    puts "=== Further instructions ==="
+
+    credentials_add_uri = ENV['TOWER_URL'].chomp('api/v1/') + '#/credentials/add/'
+    puts <<~"INFORMATION"
+      Vault credential cannot be created via API.
```
jameswnl: `vault` credential is using `kind='ssh'` with only the `vault_password` field (and `name`) filled.

Glutexo: Oh! Thanks for that information. I’ll try it out and replace the instruction with actual creation of the credential.

Glutexo:
Vault credential is now created via API, thus removed instruction to do so manually. Thanks @jameswnl  for information how to do that!

All requested changes should be resolved now. So, ready to merge? :shipit:

Glutexo:
@jameswnl Hey, just asking: Are my changes ok for this PR to be merged? It’s been hanging there for quite a long time.

Glutexo:
@jameswnl This one should be merged first. It will make possible to merge #72 as next. Can you do it, James, or should we call @Fryguy for that?

Glutexo:
Split into more atomic PRs:

* #87 Compatibility with v3.2
* #88 Missing entities
* #89 Wait for update
* #90 Stream test record

PR #79, opened 2018-05-02
[wip] Collect variable tower data

Glutexo:
Each time objects are populated in a Tower instance, some of their properties (ids, timestamps…) change. The same applies to objects counts that can differ from the previous state. This required manual update of expectation in specs when re-recording VCR cassettes.

Collect these changing data during their population in Tower instance to a YAML file. Use this information in specs instead of magic literals. This allows to re-record the cassettes without breaking the tests and removes the need to manually edit the expectations in the specs.

This PR is based on #73 and requires it to be merged before. The referenced PR fixes the specs so their expectations do not requirer other objects in Tower instance than those populated by the Rake task.

* [x] Fix new offenses detected by CodeClimate
* [x] Fix new offenses detected by MIQ Bot/RuboCop
* [ ] PR #73 is required to be merged

CC @jameswnl for review.


Glutexo:
Fixed newly introduced RuboCop and Codeclimate offenses and rebased.

Glutexo:
Broken down to smaller parts to make review easier:

* #92 Rake tasks
* #93 specs

This does not completely replace the original PR as some merges will be required. There are however only few missing bits that will come in further PRs after merging.

PR #82, opened 2018-05-10
Populate project without update job in Tower

Glutexo:
Spec examples introduced by #72 require one project with an update job and another without one. Use `another_repo` for the latter case. Alter the `populate_tower` Rake task to delete the update job after the initial update finishes.

Merging this PR does not make sense before merging other PRs:

* [ ] specs in #72 need a project without an update job

### DIFF
--- a/lib/tasks_private/spec_helper.rake
+++ b/lib/tasks_private/spec_helper.rake
@@ -40,7 +40,8 @@ class PopulateTower
   #   deleting old hello_template_with_survey: /api/v1/job_templates/590/
   # Created name=hello_template_with_survey  manager_ref/ems_ref= 593       url=/api/v1/job_templates/593/
   # created /api/v1/job_templates/594/ survey_spec
-  # Created name=another_repo                manager_ref/ems_ref= 594       url=/api/v1/projects/594/
+  # Created name=failed_repo                 manager_ref/ems_ref= 594       url=/api/v1/projects/594/
+  # Created name=jobless_repo                manager_ref/ems_ref= 595       url=/api/v1/projects/595/
   # === Object counts ===
   # configuration_script           (job_templates)     : 120
   # configuration_script_source    (projects)          : 32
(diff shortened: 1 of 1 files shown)


Glutexo:
Rebased on current state of #72.

Glutexo:
@miq-bot add_label unmergeable

Glutexo:
@miq-bot add_label pending core

Glutexo:
I rebased to the current master, so it doesn’t contain any duplicate commits. Still I’d like to note that this PR introduces changes that are only useful after #72 gets into master. Merging this PR without #72 doesn’t make much sense.

Glutexo:
The last update job is used to capture its stdout. This output is now stored only if the update fails. [\[1\]](https://github.com/ManageIQ/manageiq/pull/17290#issuecomment-398781641), [\[2\]](https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/72#issuecomment-399032381) To test this, we need more than one project: [3](https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/72/commits/f7c616e8767752551fefb78b5e75bb1fcf58b418#diff-80d88f29d306337d35e6755646d7190dR183)

* A project with successful update.
* A project with failed update.
* A project without update.

Updated the Rake task to create those projects. Removed _another_repo_, any of the new ones (_jobless_repo_ or _failed_repo_) can be used instead by https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/93 and https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/92.

tumido:
I did a cleanup on that file, it was painful to read it. I know that's not the point of this PR, though I had to do it...

tumido:
The other PR https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/72 marked as "merge before" doesn't necessarily block this one. The dependency is rather a use-case for the environment prepared by this PR. We shouldn't need to wait.

Fryguy:
In the future, please keep unrelated rubocop changes to a separate PR.

tumido:
I know. We had a discussion about that here https://github.com/ManageIQ/manageiq-providers-ansible_tower/commit/a4fe4a11ddbc2d40fbb7dea5b43b62788b5c61bf#commitcomment-29502367 so we decided to keep it this way...

tumido:
@jameswnl, can you review this once more, please? Are there any concerns which should be addressed?

PR #83, opened 2018-05-10
Check ConfigScriptSource update in provider result

Glutexo:
Send a separate notification on repository refresh in provider. Repository refresh job result does not affect create/update notification: Even if the project create/update succeeds, the refresh in provider can still fail, which is most probably not what the user expected. They can then take action when a failure notification appears.

This is loosely related to https://bugzilla.redhat.com/show_bug.cgi?id=1513616. Extracted from #72 as this is not exactly the subject of the bug/RFE.

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/shared/automation_manager/tower_api.rb
+++ b/app/models/manageiq/providers/ansible_tower/shared/automation_manager/tower_api.rb
@@ -11,9 +11,7 @@ module ManageIQ::Providers::AnsibleTower::Shared::AutomationManager::TowerApi
     def create_in_provider(manager_id, params)
       manager = ExtManagementSystem.find(manager_id)
       tower_object = raw_create_in_provider(manager, params)
-      if respond_to?(:refresh_in_provider)
-        refresh_in_provider(tower_object)
-      end
+      refresh_in_provider_notify(manager_id, params, tower_object)
       refresh(manager)
       find_by!(:manager_id => manager.id, :manager_ref => tower_object.id)
     rescue AnsibleTowerClient::ClientError, ActiveRecord::RecordNotFound => error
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/ansible_tower/shared/automation_manager/tower_api.rb
```diff
@@ -76,9 +79,12 @@ def update_in_provider(params)
     with_provider_object do |provider_object|
       provider_object.update_attributes!(params)
     end
+
     if respond_to?(:refresh_in_provider)
```
jameswnl: What do you think if we make Line79-81 as a class method, e.g. `refresh_and_notify`?  And Line 15-18 can also call this?

Glutexo: 👍  This is better than extracting just a one-liner.

It was not as straightforward

Glutexo:
As @jameswnl suggested, extracted the “if refreshable, refresh, then notify” workflow to a method.

app/models/manageiq/providers/ansible_tower/shared/automation_manager/tower_api.rb
```diff
@@ -76,9 +85,7 @@ def update_in_provider(params)
     with_provider_object do |provider_object|
       provider_object.update_attributes!(params)
     end
-    if respond_to?(:refresh_in_provider)
-      refresh_in_provider
-    end
+    self.class.send(:refresh_in_provider_notify, manager_id, params, self)
```
jameswnl: you can do `self.class.send(:refresh_in_provider_notify, manager_id, params, self.provider_object)` and so that the `refresh_in_provider_notify` parm list can be `(manager_id, params, provider_object)` and so no more checking in that method.

Glutexo: It’s not that easy, but still you’re right, James. Originally I wanted to avoid passing the Tower object, because I considered it not DRY enough and I wanted to re-use the [ConfigurationScriptSource#refresh_in_provider](https://github.com/Glutexo/manageiq-providers-ansible_tower/blob/notify_on_project_update/app/models/manageiq/providers/ansible_tower/shared/automation_manager/configuration_script_source.rb#L60) instance method. But we still use `with_provider_object` in [TowerApi#update_in_provider](https://github.com/Glutexo/manageiq-providers-ansible_tower/blob/927d85a9ce927e0cfb57e6fe528b405949f28767/app/models/manageiq/providers/ansible_tower/shared/automation_manager/tower_api.rb#L81), so we can call the [class method](https://github.com/Glutexo/manageiq-providers-ansible_tower/blob/notify_on_project_update/app/models/manageiq/providers/ansible_tower/shared/automation_manager/configuration_script_source.rb#L25) directly with both the Tower object and the model id. I am still not fully convinced it is great, but it’s less complex. And object type checks are usually ugly, so good for now.

PR #87, opened 2018-05-24
Make Populate Tower Rake task compatible with v3.2

Glutexo:
Azure (Classic) and Rackspace credential kinds are not supported by Tower 3.2. Add version conditinal to make the Rake task compatible with this recent Tower version.

Extracted from #73 for the changes to be more atomic.


Glutexo:

PR #88, opened 2018-05-24
Add mising entities from tests to Rake task

Glutexo:
The Populate Tower Rake task does not create all entities expected by specs. Add those to the task so it is possible to re-record VCRs only with the automatically populated data.

• Vault credential is required by [the `assert_credentials` method in the `refresher.rb` Spec](https://github.com/ManageIQ/manageiq-providers-ansible_tower/blob/master/spec/support/ansible_shared/automation_manager/refresher.rb#L117).
• Two projects are required by the `"will perform a targeted refresh"` example in the `refresh_configuration_script_source.rb` Specs.

Extracted from #73 for the changes to be more atomic.


Glutexo:
#89 got merged, rebased this to resolve conflicts.

PR #91, opened 2018-05-29
Drop Tower v2.x support

Glutexo:
As noted in https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/72#issuecomment-391699487, Ansible Tower v2 reached its [End of Life][1] a long time ago already. There is no need to support it by the provider. This is backed by @bascar’s and @jameswnl’s green light for the removal.

Goes together with https://github.com/ManageIQ/manageiq/pull/17490.

[1]: https://access.redhat.com/support/policy/updates/ansible-tower


Glutexo:

PR #92, opened 2018-05-29
Collect variable Tower data upon population

Glutexo:
Each time objects are populated in a Tower instance, some of their properties (ids, timestamps…) change. The same applies to objects counts that can differ from the previous state. This information is used then in the specs.

Collect these changing data during their population in Tower instance to a YAML file. This allows to fill the necessary information in the specs after re-recording the cassettes more easily. Eventually it could be used directly from the tests too.

Extracted from #79 for easier review. This part contains only changes to the Rake task. Goes along with #93.


Glutexo:
#89 got merged. Rebased this to resolve conflicts. Reused the `get_obj` method introduced by the previous PR and utilized the waiting to collect the data after they become available.

bronaghs:
@Glutexo - Can you resolve the conflicts so we can get this in

Glutexo:
Rebased on current master, conflicts resolved.

tumido:
@jameswnl, I see here that miq-bot has some suggestions which are being resolved in the https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/82 PR. Can this one be reviewed and merged. I'll resolve the eventual conflicts in the https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/82 .

lib/tasks_private/spec_helper.rake
```diff
@@ -65,42 +99,56 @@ class PopulateTower
...
   end
 
-  def create_obj(uri, data)
+  def create_obj(uri, data, dumped_fields = %w())
     del_obj(uri, data['name'])
     obj = JSON.parse(@conn.post(uri, data).body)
     puts "%s %s %s" % ["Created name=#{obj['name']}".ljust(40), "manager_ref/ems_ref= #{obj['id']}".ljust(30), "url=#{obj['url']}".ljust(10)]
+
+    dumped_data = obj.slice(*dumped_fields)
+    @tower_data.update { |td| td['items'][obj['name']] = dumped_data } unless dumped_data.empty?
```
jameswnl: Can you move the `@tower_data.update` part out of this method?  Let it just does the creation of object in Tower.

lib/tasks_private/spec_helper.rake
```diff
@@ -4,15 +4,47 @@ namespace :spec do
...
+    @data
+  end
+
+  def write
+    Spec::Support::TowerDataHelper.tower_data = @data
+  end
+
+  def update
+    yield @data if block_given?
+    write
```
jameswnl: this write each time something updated seems too much.  yeah, it's not a big deal in this rake task. but ....

lib/tasks_private/spec_helper.rake
```diff
@@ -4,15 +4,47 @@ namespace :spec do
...
   task :tower_counts do
     tower_host = ENV['TOWER_URL'] || "https://dev-ansible-tower3.example.com/api/v1/"
     id = ENV['TOWER_USER'] || 'testuser'
     password = ENV['TOWER_PASSWORD'] || 'secret'
-    PopulateTower.new(tower_host, id, password).counts
+    PopulateTower.new(tower_host, id, password, TowerData.new(false)).counts
+  end
+end
+
+class TowerData
```
jameswnl: This is pretty cool.  However, it's overkill in this case and that introduces unnecessary mental weight for forthcoming maintainers.  Can we simply dump the collected hash into yml after creating the data set?

spec/support/tower_data_helper.rb
```diff
@@ -0,0 +1,21 @@
+module Spec
```
jameswnl: Similarly here, I think we don't need this to achieve what this PR intends to. In #93 you can simply load from either  a given yml file or the default one.

tumido:
@jameswnl I think it is acceptable now. Can you please review it? Thanks :+1: :wine_glass:

lib/tasks_private/spec_helper.rake
```diff
@@ -65,16 +72,26 @@ class PopulateTower
...
+    uri = '/api/v1/me'
+    me = get_obj(uri)
+    @tower_data[:user] = { :id => me['results'].first['id'] }
   end
 
   def v3_2?
     @version >= Gem::Version.new('3.2')
   end
 
+  def save
```
jameswnl: This object is `PopulateTower` and so a method named as `save` would be ambiguous as to where it's saving to. How about rename it explicitly as `to_file`, `data_to_file` or `dump`?

lib/tasks_private/spec_helper.rake
```diff
@@ -84,24 +101,27 @@ class PopulateTower
   end
 
   def del_obj(uri, match_name)
-    data = JSON.parse(@conn.get(uri).body)
-    data['results'].each do |item|
+    obj = get_obj(uri)
+    obj['results'].each do |item|
       next if item['name'] != match_name
       puts "Deleting old #{item['name']}: #{item['url']}"
       @conn.delete(item['url'])
```
jameswnl: Found a bug not related to your changes.  We should exit this method here by `return @conn.delete(item['url'])` since we found our target. Otherwise, Line 112 will still be executed and is simply doing useless work.

lib/tasks_private/spec_helper.rake
```diff
@@ -48,13 +52,16 @@ class PopulateTower
...
+  # configuration_script_payload   (playbooks)         : 139
+  #     hello_repo                                     : 61
   #
   require 'faraday'
   require 'faraday_middleware'
 
   MAX_TRIES = 10
   TRY_SLEEP = 2
   DEL_SLEEP = 20
+  TOWER_DATA_YAML = ManageIQ::Providers::AnsibleTower::Engine.root.join('spec', 'support', 'tower_data.yml')
```
jameswnl: Remove this line and pass the file path as parameter when dumping to file (Line 11).  This way just from reading the calling code of this class, people would know what to expect.

And you don't have to split the path into 3 parts, you can simply jam them into 1 literal string being appended to the engine root.

lib/tasks_private/spec_helper.rake
tumido: I've refactored the method so it may be a bit cleaner and better to understand what it does. Tell me if you like it :wink: If not, I can revert it.

lib/tasks_private/spec_helper.rake
tumido: Sure! About the 3 parts.. I just copy-pasted that without thinking, thanks for pointing that out. :rocket:

lib/tasks_private/spec_helper.rake
```diff
@@ -4,7 +4,11 @@ namespace :spec do
     tower_host = ENV['TOWER_URL'] || "https://dev-ansible-tower3.example.com/api/v1/"
     id = ENV['TOWER_USER'] || 'testuser'
     password = ENV['TOWER_PASSWORD'] || 'secret'
-    PopulateTower.new(tower_host, id, password).create_dataset.counts
+    populate = PopulateTower.new(tower_host, id, password)
+
+    populate.create_dataset
+    populate.counts
+    populate.to_file(ManageIQ::Providers::AnsibleTower::Engine.root.join("spec/support/tower_data.yml"))
```
jameswnl: single `'` 😄

tumido: Should I open a new PR? :smile:

lib/tasks_private/spec_helper.rake
Glutexo: I don’t like the fact that the path is here in the Rake task. Since the YAML is going to be used by the Specs, I’d prefer to have the constant somewhere along them and use it from here.

lib/tasks_private/spec_helper.rake
Glutexo: I agree that this excessive writing just as the whole TowerData class might be an overkill. I’d just like to explain my motivation to that:

Because the population of a Tower instance can easily end with some failure during the process, I wanted to have the YAML file always consistent with what has been actually created in Tower. It’s always possible to revert it and it can serve as a reference.

But as @tumido pointed out, it is not actually necessary, because the YAML is intended to be used solely by Specs. And it can be used only if complete. For reference you can always read the terminal output or actually look into the Tower web UI.

lib/tasks_private/spec_helper.rake
Glutexo: I don’t like the fact that the path is here in the Rake task. Since the YAML is going to be used by the Specs, I’d prefer to have the constant somewhere along them and use it from here.

PR #93, opened 2018-05-29
Extract variable Tower data from tests

Glutexo:
Some properties of the records expected in a Tower instance are not
entered directly by the user: ids, timestamps, external data… These
are subject to change with every population and cassette re-recording.

In such case manual update of expectations in specs was
required. Collection of this data to a single resource file allows to
change only this single file instead of hunting the right magic
literals in the specs examples.

Extracted from #79 for easier review. This part contains only changes to the specs. Goes along with #92.


bronaghs:
@Glutexo - Can you address the conflicts please so we can get this in.

Glutexo:
Rebased, resolved. VCRs were re-recorded and because of that ids and counts in spec changed. If this gets merged, doing such and update would be easier for anybody. (@jameswnl in this case)

tumido:
@jameswnl, can this one please get reviewed so we can slowly get to the point when Stepan's PRs are merged? :wink:

jameswnl:
This is great!

spec/support/ansible_shared/automation_manager/refresh_configuartion_script_source.rb
```diff
@@ -14,27 +14,38 @@
   end
   let(:manager_class) { manager_class }
 
+  let(:tower_data) { Spec::Support::TowerDataHelper.tower_data }
+
+  let(:targeted_refresh_id) { tower_data['items']['targeted_refresh']['id'] }
```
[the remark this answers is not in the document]
Glutexo: This PR extracts the actual data used in the specs. I tried to map it as well as possible to the data created by the Rake task, but still they are more or less independent. If I recall correctly, the ids used here were different from the _hello_repo_ and _another_repo_, so I created a separate entity for them in the YAML.

This can be resolved when merging with the other PR by using _hello_repo_ instead and changing the data in this spec. This will however require re-recording of the VCRs, which is what I didn’t want to do here. It’s the inconsistency between specs I was talking about with you earlier. :(

spec/support/ansible_shared/automation_manager/refresh_configuartion_script_source.rb
```diff
@@ -14,27 +14,38 @@
...
+  let(:targeted_refresh_last_updated) { tower_data['items']['targeted_refresh']['last_updated'].utc }
+  let(:targeted_refresh_playbooks) { tower_data['items']['targeted_refresh']['playbooks'] }
+  let(:nonexistent_repo_id) { tower_data['items']['nonexistent_repo']['id'] }
```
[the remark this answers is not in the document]
Glutexo: As the name suggests, a _nonexistent_repo_ is an invalid id. It was used to ensure that the refresher does not even try to load this repo when updating the _targeted_refresh_ one. When merging with the Rake task changes, it would be necessary to create such id. That can be achieved by creating some repo and then deleting it, making sure the id is captured and it neither belongs to any existing project in Tower, nor going to be reused unexpectedly in the future.

lib/tasks_private/spec_helper.rake
```diff
@@ -434,6 +434,20 @@ class PopulateTower
     last_update = wait_for_project_update(jobless_project)
     @conn.delete(last_update['url'])
 
+    # Create and remove project - record an collect an ID of missing entity
+    uri = '/api/v1/projects/'
+    uri = '/api/v1/projects/'
```
Glutexo: Dupy dupy dupy dup dup dup.

Glutexo: Dupy dupy dupy dup dup dup.

lib/tasks_private/spec_helper.rake
```diff
@@ -434,6 +434,20 @@ class PopulateTower
...
+    data = {
+      :name         => 'nonexistent_repo',
+      :scm_url      => 'https://github.com/jameswnl/ansible-examples',
+      :scm_type     => 'git',
+      :credential   => scm_credential['id'],
+      :organization => organization['id']
+    }
+    nonexistent_project = create_obj(uri, data)
+    @tower_data[:items][data[:name]] = { :id => nonexistent_project['id'] }
+    del_obj(uri, data[:name])
```
Glutexo: Just remember, that `del_obj` does some safety sleep. This might be necessary when recreating object that are already present in Tower. It is however useless here, when we just want to delete an object without any intention of creating its replacement.

https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/93/commits/7f4bdfeb147d0991b05db858cb5d5a7861f05fdb#diff-bb7272be93a294c53ae6433eb024074fR108

spec/support/ansible_shared/tower_data_helper.rb
```diff
@@ -0,0 +1,3 @@
+shared_context "uses tower_data.yml" do
+  let(:tower_data) { YAML.load_file(ManageIQ::Providers::AnsibleTower::Engine.root.join("spec/support/tower_data.yml")) }
```
Glutexo: I still don’t like this. Can’t we turn the path into a constant then use it when writing the file by the Rake task?

https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/93/commits/7f4bdfeb147d0991b05db858cb5d5a7861f05fdb#diff-bb7272be93a294c53ae6433eb024074fR11

lib/tasks_private/spec_helper.rake
tumido: What's the issue with this? The rake job is run only in case of regenerating VCRs. It doesn't slow down the test run or anything...

spec/support/ansible_shared/tower_data_helper.rb
tumido: Why would be a global constant better than a shared context? The context defines the test scope. Global constant would require the file to be loaded for any spec execution within this repo. A shared context allows lazy loading of the file - only if the spec requires it. In other words, this way it is not required to have the YAML present and loaded at all when you're running a spec which doesn't include the context.

Glutexo: You just wait a _long_ time for nothing before going to the next step – populating item counts.

lib/tasks_private/spec_helper.rake
Glutexo: You just wait a *long* time for nothing before going to the next step – populating item counts.

spec/support/ansible_shared/tower_data_helper.rb
Glutexo: There is nothing bad with the shared context. I’d just extract the file path to the global content. That is virtually no overhead (as compared to loading and parsing the whole file) for that and then in can be used even outside the specs for writing the file.

tumido:
@miq-bot remove_label unmergeable

tumido:
@miq-bot add_label enhancement, test

PR #103, opened 2018-06-28
Support for Tower Workflow Job.

lfu:
Depends on https://github.com/ManageIQ/manageiq/pull/17654.
Depends on https://github.com/ansible/ansible_tower_client_ruby/pull/111.
Blocks https://github.com/ManageIQ/manageiq-ui-classic/pull/4157.

https://bugzilla.redhat.com/show_bug.cgi?id=1590975

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/automation_manager/workflow_job.rb
+++ b/app/models/manageiq/providers/ansible_tower/automation_manager/workflow_job.rb
@@ -1,8 +1,14 @@
 class ManageIQ::Providers::AnsibleTower::AutomationManager::WorkflowJob <
   ManageIQ::Providers::ExternalAutomationManager::OrchestrationStack
 
+  require_nested :Status
+
   belongs_to :ext_management_system, :foreign_key => :ems_id, :class_name => "ManageIQ::Providers::AutomationManager"
   belongs_to :workflow_template, :foreign_key => :orchestration_template_id, :class_name => "ManageIQ::Providers::AnsibleTower::AutomationManager::ConfigurationWorkflow"
+
+  def self.status_class
+    "#{name}::Status".constantize
+  end
 
   #
   # Allowed options are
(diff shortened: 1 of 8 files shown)


spec/models/manageiq/providers/ansible_tower/automation_manager/workflow_job_spec.rb
```diff
@@ -0,0 +1,3 @@
+describe ManageIQ::Providers::AnsibleTower::AutomationManager::WorkflowJob do
+  it_behaves_like 'ansible workflow job'
```
bzwei: workflow is tower only feature. Don't need to share.

app/models/manageiq/providers/ansible_tower/automation_manager/workflow_job.rb
```diff
@@ -0,0 +1,107 @@
...
+      :job_template          => job_template
+    ).tap do |stack|
+      stack.send(:update_with_provider_object, raw_job)
+      stack.update_attributes(:parent => self)
+    end
+  end
+  private :update_child_job
+
+  def update_parameters(raw_job)
+    self.parameters = YAML.safe_load(raw_job.extra_vars).collect do |para_key, para_val|
```
lfu: @jameswnl [Ansible job template job](https://github.com/ManageIQ/manageiq-providers-ansible_tower/blob/ccd0025f8b76fc79d43deffb3eb7ce6452b3ac6a/app/models/manageiq/providers/ansible_tower/shared/automation_manager/job.rb#L104) uses method ```extra_var_hash``` to convert the ```extra_var``` value from YAML format to a hash. Can we add the same method for Ansible workflow template job in to Ansible tower client?

bzwei:
@lfu can we extract common code to a mixin used by both job and workflow_job?

lfu:
Tests passed locally. A new release of Ansible tower client gem might fix it.

d-m-u:
@bzwei the two dependent PRs are merged, what else needs to happen with this one?

bzwei:
@d-m-u LGTM

gmcculloug:
@bzwei Can you look into addressing the rubocop errors.

lfu:
@gmcculloug Rubocop warnings regarding ```:inverse_of``` can't be fixed as the associations here are one way relationship.

PR #104, opened 2018-07-04
Removing old InventoryCollection definitions

slemrmartin:
Also unifying spec IC model_class testing for EmbeddedAnsible


slemrmartin:
Cc @jameswnl @agrare 
*There is cyclic dependency with core*

Ladas:
@slemrmartin hm the dependency is not merged but the CI is passing?

spec/support/ansible_shared/automation_manager/refresher.rb
```diff
@@ -237,7 +237,7 @@ def assert_inventory_root_group
     expect(expected_inventory_root_group).to have_attributes(
       :name    => "hello_inventory",
       :ems_ref => "99",
-      :type    => "ManageIQ::Providers::AutomationManager::InventoryRootGroup",
+      :type    => manager_class::InventoryRootGroup.name,
```
agrare: See https://github.com/ManageIQ/manageiq/pull/17574#issuecomment-403505086 for why I don't like just changing the spec test

slemrmartin:
@agrare now this PR doesn't depend on core

slemrmartin:
no, it has to be merged together, because deleted files here will be missing in core before core's PR will be merged too. 
https://github.com/ManageIQ/manageiq/pull/17574

PR #106, opened 2018-07-09
tower_api.update_in_provider to remove miq_task_id from params

jameswnl:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1594843

https://github.com/ManageIQ/manageiq/pull/17015 introduced this `miq_task_id` when dispatching the queue message and that is causing Tower to fail.


jameswnl:
@miq-bot add_labels bug

PR #108, opened 2018-07-16
Service AnsibleTower and EmbeddedAnsible UI parity

tumido:
Make the service layout same for AnsibleTower as it is for EmbeddedAnsible - provider part

- Extract `:raw_stdout_via_worker` to shared Ansible codebase

Related PRs: 
Fixes: [BUG 1590840](https://bugzilla.redhat.com/show_bug.cgi?id=1590840) [BUG 1590844](https://bugzilla.redhat.com/show_bug.cgi?id=1590844)

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/shared/automation_manager/job.rb
+++ b/app/models/manageiq/providers/ansible_tower/shared/automation_manager/job.rb
@@ -168,6 +168,22 @@ module ManageIQ::Providers::AnsibleTower::Shared::AutomationManager::Job
     raise MiqException::MiqOrchestrationStatusError, err.to_s, err.backtrace
   end
 
+  # Intend to be called by UI to display stdout. The stdout is stored in MiqTask#task_results or #message if error
+  # Since the task_results may contain a large block of data, it is desired to remove the task upon receiving the data
+  def raw_stdout_via_worker(userid, format = 'txt', role = nil)
+    options = {:userid => userid || 'system', :action => 'ansible_stdout'}
+    queue_options = {
+      :class_name  => self.class,
+      :method_name => 'raw_stdout',
+      :instance_id => id,
+      :args        => [format],
+      :priority    => MiqQueue::HIGH_PRIORITY,
+      :role        => role
+    }
+
+    MiqTask.generic_action_with_callback(options, queue_options)
+  end
+
   def retire_now(requester = nil)
     update_attributes(:retirement_requester => requester)
     finish_retirement
--- a/spec/models/manageiq/providers/ansible_tower/automation_manager/job_spec.rb
+++ b/spec/models/manageiq/providers/ansible_tower/automation_manager/job_spec.rb
@@ -1,3 +1,36 @@
 describe ManageIQ::Providers::AnsibleTower::AutomationManager::Job do
+  let(:job) { FactoryGirl.create(:ansible_tower_job) }
+
   it_behaves_like 'ansible job'
+
+  describe '#raw_stdout_via_worker' do
+    before do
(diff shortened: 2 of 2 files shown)


tumido:

gmcculloug:
@jameswnl Is this something you are actively reviewing?  Should anyone else be included on the review list?

spec/models/manageiq/providers/ansible_tower/automation_manager/job_spec.rb
```diff
@@ -1,3 +1,36 @@
 describe ManageIQ::Providers::AnsibleTower::AutomationManager::Job do
+  let(:job) { FactoryGirl.create(:ansible_tower_job) }
+
   it_behaves_like 'ansible job'
+
+  describe '#raw_stdout_via_worker' do
```
bzwei: can you move new spec code inside `it_hehaves_like 'ansible job'`?

app/models/manageiq/providers/ansible_tower/shared/automation_manager/job.rb
```diff
@@ -168,6 +168,22 @@ def raw_stdout(format = 'txt')
     raise MiqException::MiqOrchestrationStatusError, err.to_s, err.backtrace
   end
 
+  # Intend to be called by UI to display stdout. The stdout is stored in MiqTask#task_results or #message if error
+  # Since the task_results may contain a large block of data, it is desired to remove the task upon receiving the data
+  def raw_stdout_via_worker(userid, format = 'txt', role = nil)
```
jameswnl: This is putting on queue. So better name it something like `raw_stdout_queue` (like in [here](https://github.com/jameswnl/manageiq-providers-ansible_tower/blob/9cc3598ee1affc402f4ded2dcc6e2988ecd00064/app/models/manageiq/providers/ansible_tower/shared/automation_manager/tower_api.rb#L95-L95))

spec/models/manageiq/providers/ansible_tower/automation_manager/job_spec.rb
tumido: It won't work since this feature is also used in Embedded Ansible and in that case the task requires `embedded_ansible` role to be provided by the server. Embedded Ansible has its implementation adjusted in this PR https://github.com/ManageIQ/manageiq/pull/17712 and their very own specs are available [here](https://github.com/ManageIQ/manageiq/blob/master/spec/models/manageiq/providers/embedded_ansible/automation_manager/job_spec.rb)

In other words: If moved to the shared example these specs would fail for Embedded Ansible.

jameswnl: @tumido it's likely doable to share the spec.  Try move [this line](https://github.com/ManageIQ/manageiq/blob/master/spec/models/manageiq/providers/embedded_ansible/automation_manager/job_spec.rb#L15) to L3 and the shared specs will receive it.

tumido: @jameswnl :+1: I've moved it. Let me know if the current spec layout is acceptable. :wink: I needed to change the context flow for Embedded Ansible job_spec. Please review the core PR as well. https://github.com/ManageIQ/manageiq/pull/17712

app/models/manageiq/providers/ansible_tower/shared/automation_manager/job.rb
tumido: @jameswnl ^ I can do it, but I would change it for both. What does @bzwei think?

jameswnl:
@tumido 👍  thanks!

PR #110, opened 2018-07-19
Move spec for raw_stdout_via_worker out of shared example

bzwei:
In review comments for #108 I requested to share the spec for method `raw_stdout_via_worker`. But It cannot be shared with embedded ansible because the method has a new implementation.

This PR move the spec to tower only.


bzwei:
@miq-bot add_label bug, test
@miq-bot assign @jameswnl 
cc @tumido

bzwei:
Actually #108 and https://github.com/ManageIQ/manageiq/pull/17712 together were valid. The spec failure (reported offline) may be due to a timing between both PRs were merged.

@tumido you did good job. Thanks.

PR #111, opened 2018-07-24
Add workflow in cassettes for refresh spec

jameswnl:
- Populate `workflow` objects in target Tower and refresh spec enhanced to detect it.
- The new `tower_data.yml` is the output from the rake task. It'll be used for removing the hardcoded data in the refresh specs once https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/93 is merged


jameswnl:
@miq-bot add_labels test

PR #112, opened 2018-07-26
ConfigurationWorkflow only exists in under AnsibleTower::AutomationMa…

jameswnl:
Goes with https://github.com/ManageIQ/manageiq/pull/17720

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/automation_manager.rb
+++ b/app/models/manageiq/providers/ansible_tower/automation_manager.rb
@@ -27,6 +27,8 @@ class ManageIQ::Providers::AnsibleTower::AutomationManager < ManageIQ::Providers
   require_nested :Refresher
   require_nested :RefreshWorker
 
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/ansible_tower/automation_manager.rb
```diff
@@ -27,6 +27,8 @@ class ManageIQ::Providers::AnsibleTower::AutomationManager < ManageIQ::Providers
   require_nested :Refresher
   require_nested :RefreshWorker
 
+  has_many :configuration_workflows, :dependent => :destroy, :foreign_key => "manager_id", :inverse_of => :manager
```
jameswnl: @Ladas see this as well
This is due to the removal of the relationship from parent class https://github.com/ManageIQ/manageiq/pull/17720/files

Fryguy:
Merged core change.

PR #113, opened 2018-07-27
Move ansible spec credentials to secret.yml

tumido:
Reduced code repetitions and behaves more securely since it will strip
the credentials from VCR recordings by default

The credentials can be stored in `secrets.yml` instead of env now.
```yaml
development:
  ansible:
    userid: "YOUR_FANCY_USERNAME"
    password: "SECRET_PHRASE"
    url: "https;//here.is.my.ansible.tower.instance"
```

### DIFF
--- a/spec/factories/ext_management_system.rb
+++ b/spec/factories/ext_management_system.rb
@@ -1,0 +1,21 @@
+FactoryGirl.define do
+  trait :ansible_with_vcr_authentication do
+    zone do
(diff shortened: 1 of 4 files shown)


tumido:
@jameswnl , I was not able to deal with this part anymore, I had to fix it. I know, this is not a part of any task right now, but the code was so messy I just had to do it. :smile:

tumido:
Sorry I don't have any time to work on this anymore, if anyone is interested I'm keeping the branch around...

PR #114, opened 2018-07-27
FIX wrong VCR cassette for Embedded Ansible refresher spec

tumido:
Embedded Ansible specs in core tried to use different VCR cassette than expected, so the tests are hitting nonexistent API, therefore core specs fails.

For External Tower, rspecs look up VCRs in proper location, however when the same spec is being run from the core repo, the VCR cassette is not resolved properly.

We should consider refactoring the need to use of this helper, so the same mistake is not repeated so easily.

cc @Ladas


tumido:
@miq-bot add_label bug

tumido:
@miq-bot assing @Ladas

tumido:
@miq-bot assign @Ladas

PR #117, opened 2018-07-31
Graph refresh of configuration_scripts and configuration_workflows in one InventoryCollection

jameswnl:
Refresh will create `ConfigurationWorkflow` records but then deleting them when saving `ConfigurationScript` records.

This is because the association `:configuration_scripts` will pull in both type of records and the [saving logic will determine](https://github.com/ManageIQ/manageiq/blob/master/app/models/manager_refresh/save_collection/saver/base.rb#L123-L135) the `ConfigurationWorkflow` as to be deleted because they are not in the `InventoryCollection` that's being processed.

The fix is to have `ConfigurationScript` `InventoryCollection` to use a relationship that only collects `ConfigurationScript`.

Depends on https://github.com/ManageIQ/manageiq/pull/17785

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/automation_manager.rb
+++ b/app/models/manageiq/providers/ansible_tower/automation_manager.rb
@@ -27,8 +27,6 @@ class ManageIQ::Providers::AnsibleTower::AutomationManager < ManageIQ::Providers
   require_nested :Refresher
   require_nested :RefreshWorker
 
-  has_many :configuration_workflows, :dependent => :destroy, :foreign_key => "manager_id", :inverse_of => :manager
-
   def self.ems_type
     @ems_type ||= "ansible_tower_automation".freeze
   end
--- a/app/models/manageiq/providers/ansible_tower/inventory/persister/automation_manager.rb
+++ b/app/models/manageiq/providers/ansible_tower/inventory/persister/automation_manager.rb
@@ -3,12 +3,14 @@ class ManageIQ::Providers::AnsibleTower::Inventory::Persister::AutomationManager
 
   def initialize_inventory_collections
     %i(credentials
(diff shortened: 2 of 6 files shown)


app/models/manageiq/providers/ansible_tower/automation_manager.rb
```diff
@@ -28,6 +28,7 @@ class ManageIQ::Providers::AnsibleTower::AutomationManager < ManageIQ::Providers
   require_nested :RefreshWorker
 
   has_many :configuration_workflows, :dependent => :destroy, :foreign_key => "manager_id", :inverse_of => :manager
+  has_many :job_templates, :class_name => "ConfigurationScript", :dependent => :destroy, :foreign_key => "manager_id", :inverse_of => :manager
```
jameswnl: @Fryguy It ends up, we'll need something here for the graph refresh to work properly.  That will be used in the persister app/models/manageiq/providers/ansible_tower/inventory/persister/automation_manager.rb (in this same PR)

jameswnl:
@miq-bot add_labels bug, test

app/models/manageiq/providers/ansible_tower/automation_manager.rb
Ladas: right, so the query used in InventoryCollection can collide with others, otherwise they'll start to delete each other's records. (it's the same for old refresh in this sense)

Ladas: @jameswnl  can you print what the SQL queries look like, for ```ems.job_templates``` and ```ems.configuration_workflows```

jameswnl:
The build is broken due to the PR dependency

PR #130, opened 2018-10-05
find_by 'type' is a string, not a class

jrafanie:
Fixes a rather non-obvious error on Rails 5.1:

```
  TypeError:
    can't cast Class
```

### DIFF
--- a/spec/support/ansible_shared/automation_manager/refresher.rb
+++ b/spec/support/ansible_shared/automation_manager/refresher.rb
@@ -152,14 +152,14 @@ shared_examples_for "ansible refresher" do |ansible_provider, manager_class, ems
     expect(expected_configuration_script.authentications.count).to eq(3)
 
     # vault_credential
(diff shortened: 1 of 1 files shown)


spec/support/ansible_shared/automation_manager/refresher.rb
```diff
@@ -152,14 +152,14 @@ def assert_credentials
     expect(expected_configuration_script.authentications.count).to eq(3)
 
     # vault_credential
-    vault_credential = Authentication.all.find_by(:type => manager_class::VaultCredential, :manager_ref => '1022')
+    vault_credential = Authentication.all.find_by(:type => manager_class::VaultCredential.name, :manager_ref => '1022')
```
bdunne: Why the `.all.find_by`?

jrafanie: wow.  😖 .  I'll fix that too

PR #134, opened 2018-10-26
Tower 3.3 fix: Credential.manager_ref need to be an integer for Tower 3.3

jameswnl:
Part of the fix to https://bugzilla.redhat.com/show_bug.cgi?id=1640533

Tower 3.3 changed validation rule that now the type of reference to credentials (and others) must be an integer.

Automate code would make use of this new method

Related PR: https://github.com/ManageIQ/manageiq/pull/18154

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/shared/automation_manager/credential.rb
+++ b/app/models/manageiq/providers/ansible_tower/shared/automation_manager/credential.rb
@@ -30,7 +30,11 @@ module ManageIQ::Providers::AnsibleTower::Shared::AutomationManager::Credential
   end
 
   def provider_object(connection = nil)
-    (connection || connection_source.connect).api.credentials.find(manager_ref)
+    (connection || connection_source.connect).api.credentials.find(native_ref)
+  end
+
+  def native_ref
+    Integer(manager_ref)
   end
 
   COMMON_ATTRIBUTES = {}.freeze
--- a/spec/support/ansible_shared/automation_manager/credential.rb
+++ b/spec/support/ansible_shared/automation_manager/credential.rb
@@ -5,6 +5,20 @@ shared_examples_for "ansible credential" do
   let(:atc)           { double("AnsibleTowerClient::Connection", :api => api) }
   let(:api)           { double("AnsibleTowerClient::Api", :credentials => credentials) }
 
+  context "Tower 3.3 needs pk as integer" do
+    let(:machine_credential) { FactoryGirl.create(:ansible_machine_credential, :manager_ref => '1', :resource => manager) }
+    it "native_ref returns integer" do
+      expect(machine_credential.manager_ref).to eq('1')
+      expect(machine_credential.native_ref).to eq(1)
+    end
+
+    it "native_ref blows up for nil manager_ref" do
+      machine_credential.manager_ref = nil
+      expect(machine_credential.manager_ref).to be_nil
+      expect{ machine_credential.native_ref }.to raise_error(TypeError)
+    end
+  end
+
   context "Create through API" do
     let(:credentials)     { double("AnsibleTowerClient::Collection", :create! => credential) }
     let(:credential)      { AnsibleTowerClient::Credential.new(nil, credential_json) }


app/models/manageiq/providers/ansible_tower/shared/automation_manager/credential.rb
```diff
@@ -33,6 +33,10 @@ def provider_object(connection = nil)
     (connection || connection_source.connect).api.credentials.find(manager_ref)
   end
 
+  def manager_ref
+    self[:manager_ref].to_i
```
gmcculloug: We have tested with this change in a few scenarios and it resolves the reported issue.  👍 

@jameswnl Is there any condition you can think of where this column could end up as `nil` and we may not want to convert that to an int?

jameswnl: `nil.to_i` returns 0 and also this is implemented in `Tower` subclass.

But talked to @agrare has concern on changing the type returned.  So your team may have to explore performing the conversion in automate code

agrare: `super.to_i` would be cleaner

gmcculloug: We would have to find all the places where we do something like: https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/embedded_ansible/automation_manager/playbook.rb#L49  (There are other places that do something similar.)

Maybe the other third option is to do this at the client gem layer so the database and code can continue the same but the client knows what Tower is expecting.

jameswnl: I don't have a preference but the `self[:attr]` is based on https://stackoverflow.com/questions/21835116/how-can-i-overwrite-a-getter-method-in-an-activerecord-model

jameswnl: @gmcculloug going Tower client gem route, you mean updating https://github.com/ansible/ansible_tower_client_ruby/blob/master/lib/ansible_tower_client/base_models/job_template.rb#L12  to search/replace the `options` param? 
hmm.. I am not sure if this is a clean way ...

gmcculloug: If we want to use  something like `manager_ref_obj` can that be a virtual column?  This would still requires us to modify all the callers.

jameswnl:
@miq-bot assign @agrare

spec/support/ansible_shared/automation_manager/credential.rb
```diff
@@ -5,6 +5,15 @@
   let(:atc)           { double("AnsibleTowerClient::Connection", :api => api) }
   let(:api)           { double("AnsibleTowerClient::Api", :credentials => credentials) }
 
+  context "Tower 3.3 needs pk as integer" do
+    let(:machine_credential) { FactoryGirl.create(:ansible_machine_credential, :manager_ref => '1', :resource => manager) }
+    it "has native_ref returned as integer" do
+      expect(machine_credential.manager_ref).to eq('1')
+      d = machine_credential.native_ref
+      expect(machine_credential.native_ref).to eq(1)
+    end
```
gmcculloug: We should also add a test for when the manager_ref is `nil`.

app/models/manageiq/providers/ansible_tower/shared/automation_manager/credential.rb
```diff
@@ -33,6 +33,10 @@ def provider_object(connection = nil)
     (connection || connection_source.connect).api.credentials.find(manager_ref)
   end
 
+  def native_ref
```
bdunne: Any reason to introduce this rather than override the `manager_ref` method?

app/models/manageiq/providers/ansible_tower/shared/automation_manager/credential.rb
jameswnl: @bdunne  overriding `manager_ref` which was my initial approach.  See @agrare 's concern in this thread

app/models/manageiq/providers/ansible_tower/shared/automation_manager/credential.rb
jameswnl: That was my initial apporach. Pinged you from an outdated comment thread about that approach.

app/models/manageiq/providers/ansible_tower/shared/automation_manager/credential.rb
bdunne: I see.  I guess my concern is that all callers need to be updated and new code needs to know about the new method.

jameswnl: yeah, that's the drawback of creating a new method.

app/models/manageiq/providers/ansible_tower/shared/automation_manager/credential.rb
```diff
@@ -33,6 +33,10 @@ def provider_object(connection = nil)
     (connection || connection_source.connect).api.credentials.find(manager_ref)
```
lfu: What is expected here? ```manager_ref``` or ```native_ref```?

jameswnl: that's fine because it's being build into an url

gmcculloug: Since that is talking to the Tower API I'm pretty sure it would be `native_ref`.

jameswnl:
@agrare can we merge this?

simaishi:
Hammer backport details:

```
$ git log -1
commit a45325609bd2294ae3376d89f98154d757cd23e3
Author: Adam Grare <agrare@redhat.com>
Date:   Thu Nov 1 15:55:08 2018 -0400

    Merge pull request #134 from jameswnl/tower33
    
    Tower 3.3 fix: Credential.manager_ref need to be an integer for Tower 3.3
    
    (cherry picked from commit 73ba300544826bf50c8ca635c17a4b625349aeef)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1640533
```

PR #153, opened 2019-01-11
Targeted refresh

slemrmartin:
Targeted refresh for refreshing only specified inventory (and its subgraph) - like for amazon, azure etc.

---

* [x] depends on: https://github.com/ManageIQ/manageiq/pull/18367
* [ ] dependent: https://github.com/ManageIQ/manageiq/pull/18429
  *  needs to be merged at the same time, because embedded_ansible depends on ansible tower


app/models/manageiq/providers/ansible_tower/automation_manager/refresher.rb
```diff
@@ -1,3 +1,3 @@
-class ManageIQ::Providers::AnsibleTower::AutomationManager::Refresher < ManageIQ::Providers::BaseManager::Refresher
+class ManageIQ::Providers::AnsibleTower::AutomationManager::Refresher < ManageIQ::Providers::BaseManager::ManagerRefresher
```
agrare: @slemrmartin is this required?  I'm trying to remove ManagerRefresh in another PR

app/models/manageiq/providers/ansible_tower/shared/inventory/collector/target_collection.rb
```diff
@@ -0,0 +1,66 @@
+module ManageIQ::Providers::AnsibleTower::Shared::Inventory::Collector::TargetCollection
+  def connection
+    @connection ||= manager.connect
+  end
+
+  def initialize(_manager, _target)
+    super
+  end
```
agrare: Why is this needed if all it does is `super` ?

agrare:
@jameswnl @Ladas can you review?

app/models/manageiq/providers/ansible_tower/shared/inventory/collector/target_collection.rb
```diff
@@ -0,0 +1,66 @@
...
+    return [] if refs.blank?
+
+    if refs.size == 1
+      begin
+        [endpoint.find(refs[0])]
+      rescue AnsibleTowerClient::ResourceNotFoundError
+        [nil]
+      end
+    else
+      endpoint.all(:id__in => refs.join(','))
```
Ladas: check if there is a limit for this, e.g. for amazon, you can search for max 200 ids. Then you can do something like the multi_query here https://github.com/Ladas/manageiq-providers-amazon/blob/4197be8d1dee08a18c87483cf5ada03bebb6f5c1/app/models/manageiq/providers/amazon/inventory/collector/target_collection.rb#L23 , which is just an iterator

app/models/manageiq/providers/ansible_tower/shared/inventory/collector/target_collection.rb
slemrmartin: it's not, I'll remove it

app/models/manageiq/providers/ansible_tower/automation_manager/refresher.rb
slemrmartin: @agrare it raises error when I'm refreshing `ConfigurationScriptSource` with Refresher. 
I think there is error in `collector_inventory_for_targets()`. In `ManagerRefresher`, it derives Collector class in `Inventory`, there it tries to do by itself (https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/base_manager/refresher.rb#L141). This construction doesn't work for TargetCollection. 
Then it calls `parse_targeted_inventory` with `collector == nil`.

slemrmartin: sorry, for `InventoryRefresh::Target` it doesn't work...it contantizes `ManageIQ::Providers::AnsibleTower::Inventory::Collector::Target` instead of `ManageIQ::Providers::AnsibleTower::Inventory::Collector::TargetCollection`

slemrmartin:
@agrare @Ladas review please. 
I have to keep [wip] flag because of embedded ansible (I don't test it if it's working now - there is core dependency on this code)

app/models/manageiq/providers/ansible_tower/shared/inventory/collector/target_collection.rb
slemrmartin: it's needed now :) it adds dependent targets

app/models/manageiq/providers/ansible_tower/shared/inventory/collector/target_collection.rb
slemrmartin: @Ladas hm, I didn't find it in documentation so I need to make additional spec

spec/support/ansible_shared/automation_manager/refresh_configuartion_script_source.rb
```diff
@@ -47,7 +49,11 @@
         configuration_script_payloads = configuration_script_source.configuration_script_payloads
 
         VCR.use_cassette(cassette_path) do
-          EmsRefresh.refresh([[configuration_script_source.class.to_s, configuration_script_source.id]])
+          # EmsRefresh.refresh([[configuration_script_source.class.to_s, configuration_script_source.id]])
```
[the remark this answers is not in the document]
Ladas: So if the class wasn't listed in https://github.com/Ladas/manageiq/blob/3227d2b7f73da39db5f6b52b079977e2e32312e9/app/models/ems_refresh.rb#L136 , I suppose it never worked.

Since the target wouldn't go through ```EmsRefresh.queue_refresh```, which is used in production workflow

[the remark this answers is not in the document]
slemrmartin: anyway, this change is needed there but brokes embedded ansible specs from core (this is shared example) so there will be cycle dependency

PR #154, opened 2019-01-28
Added sleep before deleting tower's repo

slemrmartin:
Call after create and before delete of the same repo too fast can return HTTP 409.
The same sleep duration was used after delete-before create.

cc @jameswnl


lib/tasks_private/spec_helper.rake
```diff
@@ -445,6 +445,9 @@ class PopulateTower
     }
     nonexistent_project = create_obj(uri, data)
     @tower_data[:items][data[:name]] = { :id => nonexistent_project['id'] }
+
+    sleep(DEL_SLEEP)
```
jameswnl: A request: can you make the value to this `DEL_SLEEP` and 2 other parameters (Line 61) to be overridable by env variable?
These values are dependent to the setup.

slemrmartin: @jameswnl do you mean `sleep(ENV["DEL_SLEEP"] || DEL_SLEEP)` ?

jameswnl: let's do it in the declaration line: 
`DEL_SLEEP = ENV["DEL_SLEEP"] || 20`

slemrmartin: done

PR #157, opened 2019-02-07
v2 API compatibility

AlexanderZagaynov:
This change just takes into account a new credentials kind `vault` in **v2**, 
and maps it directly to already existing `VaultCredential` class.

For **v1** it still works as it was before: https://github.com/ManageIQ/manageiq-providers-ansible_tower/blob/6d7118c573636a13fc501c0e43627b9e01b73d93/app/models/manageiq/providers/ansible_tower/shared/inventory/parser/automation_manager.rb#L128-L130

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/shared/inventory/parser/automation_manager.rb
+++ b/app/models/manageiq/providers/ansible_tower/shared/inventory/parser/automation_manager.rb
@@ -118,7 +118,8 @@ module ManageIQ::Providers::AnsibleTower::Shared::Inventory::Parser::AutomationM
         'gce'        => "#{provider_module}::AutomationManager::GoogleCredential",
         'azure_rm'   => "#{provider_module}::AutomationManager::AzureCredential",
         'openstack'  => "#{provider_module}::AutomationManager::OpenstackCredential",
-        'rhv'        => "#{provider_module}::AutomationManager::RhvCredential"
+        'rhv'        => "#{provider_module}::AutomationManager::RhvCredential",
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/ansible_tower/shared/inventory/parser/automation_manager.rb
```diff
@@ -120,16 +120,22 @@ def miq_credential_types
   def credentials
     provider_module = ManageIQ::Providers::Inflector.provider_module(collector.manager.class).name
     collector.credentials.each do |credential|
+      credential_kind = credential.kind
```
jameswnl: I don't quite see what is effectively changed in this file

jameswnl:
does this PR make the refresh working?

app/models/manageiq/providers/ansible_tower/shared/inventory/parser/automation_manager.rb
AlexanderZagaynov: leftovers of previous approach

AlexanderZagaynov:
@jameswnl it is not by itself, but together with client gem.

However, current PR can be merged without waiting, 
because it doesn't change previous behavior, just extends it.

AlexanderZagaynov:
@jameswnl I believe this one can be merged without waiting. What do you think?

mfeifer:
@jameswnl see Alex's comment above.

AlexanderZagaynov:
this PR is a complement for https://github.com/ansible/ansible_tower_client_ruby/pull/123

PR #165, opened 2019-03-21
Update cassettes since client eliminated redirects

jrafanie:
Collections and many other requests now request a single `api/v1/project/`
instead of `api/v1/project` and redirecting to `api/v1/project/`.  Cassettes
were updated to remove the no longer redirected request since it no longer happens.

The client updated to remove redirects here:
ansible/ansible_tower_client_ruby#124


manageiq-providers-ansible_tower.gemspec
```diff
@@ -14,6 +14,7 @@ Gem::Specification.new do |s|
   s.files = Dir["{app,config,lib}/**/*"]
 
   s.add_runtime_dependency "ansible_tower_client", "~> 0.19"
+  # s.add_runtime_dependency "ansible_tower_client", "~> 0.20"
```
jrafanie: This PR will fail until the new client is released containing the changes in https://github.com/ansible/ansible_tower_client_ruby/pull/124

jrafanie:
@AlexanderZagaynov I was able to update the cassettes for the changes you made to eliminate the redirects.  Once we get that gem released, I can update this one to use it and should be green on both ruby 2.4 and 2.5.

AlexanderZagaynov:
@jrafanie thank you! I've prepared a follow-up PR with one more redirect fixed: https://github.com/ansible/ansible_tower_client_ruby/pull/127

jrafanie:
@miq-bot rm_label wip

agrare:
@jrafanie were some of those vcrs just not used?

jrafanie:
> @jrafanie were some of those vcrs just not used?

@agrare No, I believe all of them are used.  I changed the cassettes that we had test failures for.  Once in the cassette, I removed the whole request that was showing 301/redirect (and the test was complaining that wasn't be requested) but kept the followup request that was valid and returned a 200.

jrafanie:
@agrare hold on, let me verify something before you merge...

agrare:
Ohh okay, I saw only deletes on `pec/vcr_cassettes/manageiq/providers/ansible_tower/automation_manager/event_catcher/stream.yml` and it looked like the whole file was removed

jrafanie:
> Ohh okay, I saw only deletes on `pec/vcr_cassettes/manageiq/providers/ansible_tower/automation_manager/event_catcher/stream.yml` and it looked like the whole file was removed

Ah, if you expand the diff for that file, you'll see there is still some content in the cassettes.  Before there was a request for `something` which had a 301 code followed by a `something/` request that returned a 200.  We no longer do the first request without the trailing slash/301 response so I removed just these.  The `something/` request still immediately follow any lines I deleted.

Good call though, there was so many more deletions to be done.  This should be ready to go now.

manageiq-providers-ansible_tower.gemspec
```diff
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
 
   s.files = Dir["{app,config,lib}/**/*"]
 
-  s.add_runtime_dependency "ansible_tower_client", "~> 0.19"
+  s.add_runtime_dependency "ansible_tower_client", ">= 0.19.1"
```
jrafanie: This change is worth a comment... It didn't seem wise to use `~> 0.19.1` since that would allow only patch updates and it seemed like a lot of effort to do `~> 0.19, >= 0.19`, so I just changed it to allow anything greater than 0.19.1.  We can always be more cautious if we needed to be in the future.

jrafanie:
@agrare I think this will manageiq master builds will fail until this is merged since the cassettes are expecting redirects but not getting them when they pull down the 0.19.1 ruby client.

simaishi:
@jrafanie @agrare Travis is now failing in Hammer/Gaprindashvili branches as it's installing `ansible_tower_client` 0.19.1. I see this PR fixed master, but we don't have a reason to bump the gem version in H/G branches. Not sure if proper fix on those branches will be to lock to 0.19.0??  I'd appreciate if you can take a look. Thanks!

jrafanie:
@agrare I can go either way on this one.  0.19.1 was a bug fix release to remove unnecessary requests that would 301/redirect and the double trailing slash in the URI.  I see no reason why this PR can't be backported.  thoughts?

agrare:
:+1: I'm good with backporting this

simaishi:
@dmetzger57 `ansible_tower_client` 0.19.1 isn't fixing any BZs targeted for `hammer` branch. Is it ok to bump, or prefer to lock to `0.19.0`?

dmetzger57:
👍 I'm good with bumping

simaishi:
Backported to Hammer via https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/167

simaishi:
@jrafanie @agrare For `gaprindashvili` branch... we're at `0.17.0` right now (has `~>0.17` in [Gemfile](https://github.com/ManageIQ/manageiq/blob/gaprindashvili/Gemfile#L88)). I'm thinking we should lock to 0.17 rather than bumping to `0.19.1`. What do you think?

jrafanie:
I agree.  Without looking at the changes in those versions, it seems too risky to upgrade from 0.17.0 to 0.19.1

PR #170, opened 2019-04-29
Improvement to Ansible Tower template runner.

lfu:
Blocks https://github.com/ManageIQ/manageiq-content/pull/528.

https://bugzilla.redhat.com/show_bug.cgi?id=1623636

@miq-bot add_label enhancement, hammer/no, changelog/yes

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/automation_manager/template_runner.rb
+++ b/app/models/manageiq/providers/ansible_tower/automation_manager/template_runner.rb
@@ -15,11 +15,11 @@ class ManageIQ::Providers::AnsibleTower::AutomationManager::TemplateRunner < ::J
       (options[:execution_ttl].present? ? options[:execution_ttl].try(:to_i) : DEFAULT_EXECUTION_TTL) * 60
   end
 
-  def start
+  def start(priority: nil)
     time = Time.zone.now
     update_attributes(:started_on => time)
     miq_task.update_attributes(:started_on => time)
-    my_signal(false, :launch_ansible_tower_job, :priority => MiqQueue::HIGH_PRIORITY)
+    my_signal(false, :launch_ansible_tower_job, :priority => priority)
   end
 
   def launch_ansible_tower_job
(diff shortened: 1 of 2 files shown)


lfu:
@miq-bot add_reviewer @bzwei

app/models/manageiq/providers/ansible_tower/automation_manager/template_runner.rb
```diff
@@ -19,7 +19,7 @@ def start
     time = Time.zone.now
     update_attributes(:started_on => time)
     miq_task.update_attributes(:started_on => time)
-    my_signal(false, :launch_ansible_tower_job)
+    my_signal(false, :launch_ansible_tower_job, :priority => MiqQueue::HIGH_PRIORITY)
```
bzwei: can we default to normal priority, and accept an option to increase priority?

bzwei: Also while the start may need to be in high priority (caller is waiting to get a task handle), the retry/wait_for_completion can in normal priority.

lfu: That is how it is implemented.

app/models/manageiq/providers/ansible_tower/automation_manager/template_runner.rb
```diff
@@ -102,16 +113,16 @@ def load_transitions
...
     else
-      queue_signal(action, *args, :deliver_on => deliver_on)
+      queue_signal(action, *args, :deliver_on => deliver_on, :priority => priority)
     end
   end
 
-  def queue_signal(*args, deliver_on: nil)
-    priority = options[:priority] || MiqQueue::NORMAL_PRIORITY
+  def queue_signal(*args, deliver_on: nil, priority: nil)
+    priority ||= options[:priority] || MiqQueue::NORMAL_PRIORITY
```
lfu: @bzwei The default priority is normal.

app/models/manageiq/providers/ansible_tower/automation_manager/template_runner.rb
bzwei: I mean let the caller to decide the priority when call `start` or `create_job`, not hardcode the `HIGH_PRIORITY` here.

bzwei:
@lfu Please fix CI build failure

PR #174, opened 2019-06-06
Fix attribute_changed? deprecation for rails 5.1

jrafanie:
- [x] Depended on https://github.com/ManageIQ/manageiq/pull/18076 [DONE]

See also:
https://www.ombulabs.com/blog/rails/upgrades/active-record-5-1-api-changes.html
```
DEPRECATION WARNING: The behavior of `attribute_changed?` inside of after callbacks will be changing in the next version of Rails. The new return value will reflect the behavior of calling the method after `save` returned (e.g. the opposite of what it returns now). To maintain the current behavior, use `saved_change_to_attribute?` instead. (called from change_maintenance_for_provider at /Users/joerafaniello/Code/manageiq-providers-ansible_tower/app/models/manageiq/providers/ansible_tower/shared/automation_manager.rb:31)
.DEPRECATION WARNING: The behavior of `attribute_changed?` inside of after callbacks will be changing in the next version of Rails. The new return value will reflect the behavior of calling the method after `save` returned (e.g. the opposite of what it returns now). To maintain the current behavior, use `saved_change_to_attribute?` instead. (called from block in included at /Users/joerafaniello/Code/manageiq-providers-ansible_tower/app/models/manageiq/providers/ansible_tower/shared/automation_manager.rb:15)
```

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/shared/automation_manager.rb
+++ b/app/models/manageiq/providers/ansible_tower/shared/automation_manager.rb
@@ -12,7 +12,7 @@ module ManageIQ::Providers::AnsibleTower::Shared::AutomationManager
            :to => :provider
 
   def self.included(klass)
-    klass.after_save :change_maintenance_for_provider, :if => proc { |ems| ems.enabled_changed? }
+    klass.after_save :change_maintenance_for_provider, :if => proc { |ems| ems.saved_change_to_enabled? }
   end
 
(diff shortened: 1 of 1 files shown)


jrafanie:
Note, this won't pass tests until rails 5.1 is merged on manageiq

jrafanie:
@miq-bot rm_label wip

PR #187, opened 2019-09-13
Unsharing previously shared ansible tower code

jameswnl:
Followup to https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/186


jameswnl:
@miq-bot add_labels cleanup

PR #189, opened 2019-09-16
Unshare inventory collector target_collection

jameswnl:
Found an issue with existing specs: it has been `stud_const` the wrong `"#{manager_class.parent}::Inventory::Collector::TargetCollection::MAX_FILTER_SIZE"` and so have been testing the same `batch_size` (200) for the 3 iterations.

This PR will just focus on unsharing the code and hence modify the the spec a bit to "maintain" its uselessness.  Will have to re-record the cassettes and be dealt with in separate issue/PR

PR #190, opened 2019-09-16
fix some specs to make them meaningful

jameswnl:
The issue was discovered when working on https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/189

Now fixed and cassette re-recorded.


jameswnl:
@miq-bot add_labels bug

jameswnl:
@miq-bot add_labels test

PR #193, opened 2019-09-18
[event_parser.rb] Filter out extra_vars data

NickLaMuro:
This is an alternative to https://github.com/ManageIQ/manageiq/pull/19308

Since we can't be sure if there is sensitive data in there, it is better to just filter this out in the logs.  Hopefully this data isn't being used anywhere when parsing `job_create` events...

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/automation_manager/event_parser.rb
+++ b/app/models/manageiq/providers/ansible_tower/automation_manager/event_parser.rb
@@ -4,11 +4,17 @@ module ManageIQ::Providers::AnsibleTower::AutomationManager::EventParser
   end
 
   def event_to_hash(event, ems_id)
+    filtered_event_data = event.dup.tap do |data|
+      if (changes_hash = data[:full_data]["changes"])
+        changes_hash["extra_vars"] = '[FILTERED]' if changes_hash["extra_vars"]
+      end
+    end
+
     {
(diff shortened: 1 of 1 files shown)


Fryguy:
cc @dmetzger57 @tinaafitz

Fryguy:
Note that if we backport this, we *may* need a second identical fix to the old tower-based embedded ansible.

NickLaMuro:
@Fryguy Yes, we would need a `hammer`  specific one, but that would only be making the same change in what used to be the `app/models/manageiq/providers/ansible-tower/shared` directory.  There shouldn't have to be a core PR for `hammer`.

NickLaMuro:
@miq-bot remove label hammer/yes

Hammer is being backported here:  https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/194

NickLaMuro:
bah...

@miq-bot remove_label hammer/yes

simaishi:
Backported to hammer via https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/194

simaishi:
@NickLaMuro backporting to `ivanchuk` conflicts as it's using shared parser, maybe same conflict as hammer... anyway, please create a separate PR for ivanchuk branch.

NickLaMuro:
@simaishi I created a ivanchuk branch and assigned you that PR.  Let me know if you need anything else.  Thanks!

PR #194, opened 2019-09-24
[HAMMER][event_parser.rb] Filter out extra_vars data

NickLaMuro:
Since we can't be sure if there is sensitive data in there, it is better to just filter this out in the logs.  Hopefully this data isn't being used anywhere when parsing `job_create` events...

This is the `hammer` backport of the following PR:

https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/193

With the proper code changes done the first time...

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1758665

PR #195, opened 2019-09-24
[event_parser.rb] Properly filter events

NickLaMuro:
Now with 100% moar bettah fixes!

Description
-----------

`:full_data` isn't a key that exists in the original `event` hash, but is the key that we are setting the entirety of the `event` has too.

This blunder was caused by a fat fingered port of the code from this core PR:

https://github.com/ManageIQ/manageiq/pull/19308

And I forgot to remove this other bit when converting the `filtered_event_data` for this fix... sorry...

Links
-----

* https://bugzilla.redhat.com/show_bug.cgi?id=1752033


NickLaMuro:
@miq-bot add_label ivanchuk/yes

simaishi:
Backported to hammer via https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/194

PR #198, opened 2019-09-30
Add DDF based verify_credentials

Fryguy:
Part of https://github.com/ManageIQ/manageiq/issues/18818

Completely untested, but since this is my first, want to get some 🍅 thrown at it. @agrare Please review.

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/provider.rb
+++ b/app/models/manageiq/providers/ansible_tower/provider.rb
@@ -8,6 +8,88 @@ class ManageIQ::Providers::AnsibleTower::Provider < ::Provider
   validates :name, :presence => true, :uniqueness => true
   validates :url,  :presence => true
 
+  PARAMS_FOR_CREATE = {
+    :title  => "Configure Ansible Tower",
+    :fields => [
+      {
+        :component  => "text-field",
+        :name       => "endpoints.default.base_url",
+        :label      => "URL",
+        :isRequired => true,
+        :validate   => [{:type => "required-validator"}]
+      },
+      {
+        :component  => "text-field",
+        :name       => "endpoints.default.username",
+        :label      => "Username",
+        :isRequired => true,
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/ansible_tower/provider.rb
```diff
@@ -8,6 +8,75 @@ class ManageIQ::Providers::AnsibleTower::Provider < ::Provider
...
+  #   }
+  # }
+  def self.verify_credentials(args)
+    default_endpoint = args.dig("endpoints", "default")
+
+    base_url, username, password, verify_ssl = default_endpoint&.values_at(
+      "base_url", "username", "password", "verify_ssl"
+    )
+
+    !!verify_connection(raw_connect(base_url, username, password, verify_ssl))
```
agrare: AnsibleTowerClient expects verify_ssl as VERIFY_NONE/VERIFY_PEER not true, false [ref](https://github.com/ansible/ansible_tower_client_ruby/blob/master/lib/ansible_tower_client/connection.rb#L11-L12)
`verify_ssl = verify_ssl ? OpenSSL::SSL::VERIFY_PEER : OpenSSL::SSL::VERIFY_NONE` ?

agrare:
Just the verify_ssl issue, then this works in the success and failure case for both the provider and automation_manager :+1:

agrare:
Also the base_url wouldn't work unless it was in the `https://.../api/v1` format, we might want to move the url formatting from [`#url=`](https://github.com/ManageIQ/manageiq-providers-ansible_tower/blob/1e457fb6d13168d8603d94e735c15eb8d9ec85c0/app/models/manageiq/providers/ansible_tower/provider.rb#L115-L120) up to the class level so the UI doesn't have to do this before passing the base_url in

Fryguy:
@agrare Updated with the base_url handling

PR #202, opened 2019-10-29
Changed event_parser from module to class

juliancheal:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1650457

Seems that in commit https://github.com/ManageIQ/manageiq-providers-ansible_tower/commit/43f5c26989b97ee6eb407b357dc4937195808b8b

We changed event_parser.rb to extending ManageIQ::Providers::AnsibleTower::Shared::AutomationManager::EventParser

However now it as a module we would need to add self to event_to_hash method.

Changing EventParser to a class, as it no longer needs to be a module


juliancheal:
@NickLaMuro Just wanted you to check this is what we discussed on gitter.

agrare:
This is a good fix but I don't believe it is the cause of that BZ because this change wasn't backported to 5.10

PR #210, opened 2020-03-05
Address strict partial failures by using explicit Ansible provider factories

djberg96:
Enabling strict partials revealed several instances of this error:

```
Failure/Error: allow_any_instance_of(Provider).to receive_messages(:connect => connection)
       Provider does not implement #connect
```
Currently the ems provider is just `Provider` in many cases, when it should be a fully scoped `ManageIQ::Providers::AnsibleTower::Provider`. Since the base `Provider` doesn't implement a `connect` method, the specs fail in strict mode.

This PR cleans up most of the current issues, mainly by creating and setting an explicit ansible provider (which does implement a `connect` method, and presumably is what we want to actually use anyway), as well as updating the partial in the various `allow_any_instance_of` calls. There was also a similar issue with the configuration script model.

### DIFF
--- a/spec/models/manageiq/providers/ansible_tower/automation_manager/configuration_script_spec.rb
+++ b/spec/models/manageiq/providers/ansible_tower/automation_manager/configuration_script_spec.rb
@@ -1,4 +1,4 @@
-describe ManageIQ::Providers::AnsibleTower::AutomationManager::ConfigurationScript do  
+describe ManageIQ::Providers::AnsibleTower::AutomationManager::ConfigurationScript do
   it 'designates orchestration stack type' do
     expect(described_class.stack_type).to eq('Job')
   end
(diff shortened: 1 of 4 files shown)


djberg96:
@miq-bot add_label test, refactoring

spec/models/manageiq/providers/ansible_tower/automation_manager/configuration_script_spec.rb
```diff
@@ -13,9 +13,13 @@
   end
 
   context "ansible configuration_script" do
-    let(:provider_with_authentication)       { FactoryBot.create(:provider_ansible_tower, :with_authentication) }
-    let(:manager_with_authentication)        { provider_with_authentication.managers.first }
-    let(:manager_with_configuration_scripts) { FactoryBot.create(:automation_manager_ansible_tower, :provider, :configuration_script) }
+    let(:provider_with_authentication) { FactoryBot.create(:provider_ansible_tower, :with_authentication) }
+    let(:manager_with_authentication)  { provider_with_authentication.managers.first }
```
agrare: > `provider_with_authentication.managers.first`

This could be `provider_with_authentication.automation_manager`

djberg96: @agrare Yes, that's nicer, updated.

PR #211, opened 2020-03-05
Set EMS#enabled in Provider#ensure_managers

agrare:
Auto-creating an automation_manager when creating the parent Provider fails when the provider is first created in the maintenance zone.  This is because the automation_manager that gets created has the provider's zone.id but not the enabled column set.

`Validation failed: ManageIQ::Providers::AnsibleTower::Provider: Zone cannot be the maintenance zone when provider is active`

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/provider.rb
+++ b/app/models/manageiq/providers/ansible_tower/provider.rb
@@ -134,6 +134,9 @@ class ManageIQ::Providers::AnsibleTower::Provider < ::Provider
   def ensure_managers
     build_automation_manager unless automation_manager
     automation_manager.name    = _("%{name} Automation Manager") % {:name => name}
-    automation_manager.zone_id = zone_id if zone_id_changed?
+    if zone_id_changed?
+      automation_manager.enabled = Zone.maintenance_zone&.id != zone_id
+      automation_manager.zone_id = zone_id
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


djberg96:
@agrare There's a factory on line 2 in `spec/support/ansible_shared/automation_manager.rb` that I think we should update as well.

bdunne:
@djberg96 Are you applying this change to the rest of the plugins that define `ensure_managers`?

djberg96:
@bdunne I wasn't planning on it. Should we?

bdunne:
Any reason not to?  It seems like a valid issue across the board

djberg96:
@bdunne You tell me. I guess I wasn't aware it was an issue in other providers, I was just following the spec failures. If so, we should probably collect a list of affected providers, create an issue, and tick them off as completed.

agrare:
I believe these are the only ones affected:
```
./manageiq/app/models/manageiq/providers/embedded_ansible/provider.rb
./manageiq-providers-foreman/app/models/manageiq/providers/foreman/provider.rb
```

None of the other `Providers` auto-creates child managers.

bdunne:
I know this same method exists in both core and manageiq-providers-foreman

Fryguy:
I thought OpenStack has it as well?

bdunne:
https://github.com/ManageIQ/manageiq/blob/09cbc06f78c399eee1a0bda1fd233bcf9c435512/app/models/mixins/has_network_manager_mixin.rb#L26

agrare:
Specifically the issue of creating a provider in maintenance mode, that is a parent ems and child ems's.

agrare:
Openstack does have a provider that includes the cloud_manager and infra_manager but it doesn't auto-create/ensure_managers

PR #219, opened 2020-05-01
Added raw_artifacts to Job

gubbe505:
Provided the Ansible job to return the raw_artifacts set by a playbook using set_stats

Ansible playbook
```yaml
  - name: Return variables
    set_stats:
      data:
        key: [value]
      aggregate: false
```

job.raw_artifact
```
{
  "key": value
}
```

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/automation_manager/job.rb
+++ b/app/models/manageiq/providers/ansible_tower/automation_manager/job.rb
@@ -169,6 +169,18 @@ class ManageIQ::Providers::AnsibleTower::AutomationManager::Job <
     raise MiqException::MiqOrchestrationStatusError, err.to_s, err.backtrace
   end
 
+  def raw_artifacts
+    ext_management_system.with_provider_connection do |connection|
+      connection.api.jobs.find(ems_ref).artifacts
+    end
+  rescue AnsibleTowerClient::ResourceNotFoundError
+    msg = "AnsibleTower Job #{name} with id(#{id}) or its artifacts does not exist on #{ext_management_system.name}"
+    raise MiqException::MiqOrchestrationStackNotExistError, msg
+  rescue => err
+    _log.error "Reading AnsibleTower Job #{name} with id(#{id}) artifacts failed with error: #{err}"
+    raise MiqException::MiqOrchestrationStatusError, err.to_s, err.backtrace
+  end
+
   # Intend to be called by UI to display stdout. The stdout is stored in MiqTask#task_results or #message if error
   # Since the task_results may contain a large block of data, it is desired to remove the task upon receiving the data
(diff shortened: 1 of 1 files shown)


gtanzillo:
@gubbe505 Please add some tests for this change. 

Also, I'm not very familiar with `raw_artficats` in ansible jobs. Do you have some examples you can share? And, lastly, do you have an idea of how much data these changes could generate?

gubbe505:
@gtanzillo 
In a ansible playbook we use the set_fact to pass on variables, like computed value (generated id). 
https://docs.ansible.com/ansible/latest/modules/set_fact_module.html

These facts are also return by a ansible job after it is finished and can be pick up by MIQ just like the stdout in the raw_output. 

The problem with the raw_output is that we have to scrape the lines to get the value that we seek. By using the artifacts the data is return is a structured variable. As single string or a complex structure.

Example playbook
```
- name: Generate Unique ID
  set_fact:
    unique_id: "{{ 999999999999 | random | to_uuid | upper }}"
```

Example manageiq code
`id = ansible_job.raw_artifacts['unique_id']`

I don't believe it will generate must extra data as it is a existing part of a ansible job.

gubbe505:
2de example 

playbook
```
- name: Return variables
  set_stats:
    data:
      osp_ipv4_segment_id: "{{ segment_id }}"
      osp_ipv4_cidr: "{{ cidr }}"
      osp_ipv4_subnet_id: "{{ subnet_id }}"
      osp_ipv4_gateway_address: "{{ gateway_address }}"
    aggregate: false 
```

manageiq code
segment_id = ansible_job.raw_artifacts['osp_ipv4_segment_id']
gateway_address = ansible_job.raw_artifacts['osp_ipv4_gateway_address']

NickLaMuro:
Removing myself from assignment on this one since I lack the context needed to be able to properly review this one from a provider perspective and ansible tower perspective.  I am pretty familiar with the inner workings of AWX, but only in relation to what we implemented for `EmbeddedAnsible` in core, and don't think I would be able to provide anything useful beyond what the requested reviewers (Adam and Jason) would.

app/models/manageiq/providers/ansible_tower/automation_manager/job.rb
```diff
@@ -169,6 +169,18 @@ def raw_stdout(format = 'txt')
     raise MiqException::MiqOrchestrationStatusError, err.to_s, err.backtrace
   end
 
+  def raw_artifacts
+    ext_management_system.with_provider_connection do |connection|
+      connection.api.jobs.find(ems_ref).artifacts
+    end
```
agrare, line 175: Future refactoring but most of these methods do "with_provider_connection { |c| c.api.jobs.find(ems_ref) }` which is essentially what `with_provider_object { |job| }` does.

PR #229, opened 2020-07-22
Delegate authentications= and endpoints= to provider

agrare:
For the create_from_params and edit_with_params methods to delegate these to the provider we have to delegate the setters as well.

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/automation_manager.rb
+++ b/app/models/manageiq/providers/ansible_tower/automation_manager.rb
@@ -61,10 +61,7 @@ class ManageIQ::Providers::AnsibleTower::AutomationManager < ManageIQ::Providers
   end
 
   def change_maintenance_for_provider
-    if provider.present? && saved_change_to_zone_id?
-      provider.zone_id = zone_id
-      provider.save
-    end
+    provider.save
   end
 
   def self.ems_type
--- a/app/models/manageiq/providers/ansible_tower/automation_manager.rb
+++ b/app/models/manageiq/providers/ansible_tower/automation_manager.rb
@@ -80,6 +80,32 @@ class ManageIQ::Providers::AnsibleTower::AutomationManager < ManageIQ::Providers
     n_('Automation Manager (Ansible Tower)', 'Automation Managers (Ansible Tower)', number)
   end
 
+  def self.create_from_params(params, endpoints, authentications)
+    new(params).tap do |ems|
+      endpoints.each { |endpoint| ems.assign_nested_endpoint(endpoint) }
+      authentications.each { |authentication| ems.assign_nested_authentication(authentication) }
+
+      ems.provider.save!
+      ems.save!
+    end
+  end
+
+  def edit_with_params(params, endpoints, authentications)
+    tap do |ems|
+      transaction do
+        # Remove endpoints/attributes that are not arriving in the arguments above
+        ems.endpoints.where.not(:role => nil).where.not(:role => endpoints.map { |ep| ep['role'] }).delete_all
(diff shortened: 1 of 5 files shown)


agrare:
cc @skateman

skateman:
I like this @agrare let me know when it is in a shape for testing in the UI :wink:

agrare:
All green taking out of WIP

skateman:
Changing the name of a provider doesn't work for me :confused: it stays the same...

agrare:
Weird that's one of the test cases, thought I was going to be able to not have to override `create_with_params` like foreman since it was passing but look like no such luck

Try it now

app/models/manageiq/providers/ansible_tower/automation_manager.rb
```diff
@@ -8,14 +8,20 @@ class << self
...
            :connect,
+           :endpoints,
+           :endpoints=,
+           :url,
+           :url=,
            :verify_credentials,
            :with_provider_connection,
            :to => :provider
 
+  belongs_to :provider, :autosave => true, :dependent => :destroy
```
gtanzillo, line 24: Was or is there a `belongs_to :provider` somewhere else? Trying to understand how the existing delegations were working.

agrare: Yes there is in core but we need the autosave and dependent destroy

himdel:
Looks like this is causing ui travis failure...

https://travis-ci.com/github/ManageIQ/manageiq-ui-classic/jobs/364237496#L2495

```
  1) AutomationManagerController#edit should save the zone field
     Failure/Error: @provider.name       = params[:name]
     
     NoMethodError:
       undefined method `sub' for nil:NilClass
     # ./app/controllers/mixins/automation_manager_controller_mixin.rb:178:in `sync_form_to_instance'
     # ./app/controllers/mixins/manager_controller_mixin.rb:17:in `add_provider'
     # ./app/controllers/mixins/manager_controller_mixin.rb:108:in `edit'
     # ./spec/controllers/automation_manager_controller_spec.rb:150:in `block (3 levels) in <top (required)>'
```

(`@provider` is `#<ManageIQ::Providers::AnsibleTower::Provider id: 25000000000628, type: "ManageIQ::Providers::AnsibleTower::Provider", name: "provider_0000000000087", guid: "26e5c619-6729-4d82-86b0-f4d05620b682", zone_id: 25000000000226, created_at: "2020-07-23 22:18:56", updated_at: "2020-07-23 22:18:56", tenant_id: 25000000000205>`)

looks like this is only happening when `params[:name]` is nil, so maybe we should just fix the test, WDYT?

app/models/manageiq/providers/ansible_tower/provider.rb
```diff
@@ -153,14 +156,18 @@ def url=(new_url)
     default_endpoint.url = self.class.adjust_url(new_url).to_s
   end
 
+  def name=(val)
+    super(val.sub(/ Automation Manager$/, ''))
```
himdel, line 160: blowing up here, maybe there should be a `to_s`?

agrare:
Ah yeah name is required so it would just fail later on in a presence validator so probably best to fix the test but also handle the nil case here so it has a better exception

himdel:
Thanks, fixing the UI part in https://github.com/ManageIQ/manageiq-ui-classic/pull/7233 :)

agrare:
Awesome thanks, and here is the provider side fix https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/231

PR #249, opened 2021-01-15
Move logger initialization and configuration into plugins

Fryguy:
Move logger initialization and configuration into plugins


Fryguy:
Oops - double blasted - prefer #250

PR #250, opened 2021-01-15
Move logger initialization and configuration into plugins

Fryguy:
Move logger initialization and configuration into plugins


agrare (approved):
Used by `AnsibleTowerClient.logger = $ansible_tower_log` :+1:

PR #253, opened 2021-02-12
Convert `supports_#{op}?` to `supports?(op)`

chessbyte:
Additionally, convert method definitions such as

```ruby
def supports_port?
  true
end
```

to

```ruby
supports :port
```

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/automation_manager/configuration_script.rb
+++ b/app/models/manageiq/providers/ansible_tower/automation_manager/configuration_script.rb
@@ -20,9 +20,7 @@ class ManageIQ::Providers::AnsibleTower::AutomationManager::ConfigurationScript 
     "Job"
   end
 
-  def supports_limit?
-    true
-  end
+  supports :limit
 
   def self.provider_collection(manager)
     manager.with_provider_connection do |connection|
--- a/app/models/manageiq/providers/ansible_tower/automation_manager/configuration_workflow.rb
+++ b/app/models/manageiq/providers/ansible_tower/automation_manager/configuration_workflow.rb
@@ -14,9 +14,7 @@ class ManageIQ::Providers::AnsibleTower::AutomationManager::ConfigurationWorkflo
     "WorkflowJob"
   end
 
-  def supports_limit?
-    false
-  end
+  supports :limit
 
   def self.display_name(number = 1)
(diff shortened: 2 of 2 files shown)


app/models/manageiq/providers/ansible_tower/automation_manager/configuration_script.rb
```diff
@@ -20,9 +20,7 @@ def self.stack_type
     "Job"
   end
 
-  def supports_limit?
-    true
-  end
+  supports :limit
```
agrare, line 23: @chessbyte `limit` doesn't appear to be a valid supports feature.  It is a little generic, wonder if we should call this something more helpful

chessbyte: Any suggestions?  We really should get an understanding of what this does based on the callers, I guess.

Fryguy: limit is the ansible feature - it's a way to further restrict playbooks to specific hosts.  So if a playbook is declared `hosts: all`, you can run it with `--limit datacenter2` to restrict it to just that host.

https://docs.ansible.com/ansible/latest/user_guide/intro_patterns.html#patterns-and-ansible-playbook-flags

Fryguy: So maybe `supports :hosts_limit` ?

Fryguy: actually, we don't use the terms hosts in configuration manager...Not sure what the alternative term is offhand (configured_systems?), but we may want to to be that `<that-term>_limit`

agrare: If we can scope the feature to just `ConfigurationScript` like @chessbyte's supports feature mixin refactoring PR I'd be good with `hosts_limit`.

Having that in the main QUERYABLE_FEATURES hash it would be too generic IMO

chessbyte: I am envisioning a world where the only globally supported features are CRUD+refresh - or maybe even less than that ;-)

chessbyte:
Closing this PR.  Will return to it after https://github.com/ManageIQ/manageiq/pull/21046 gets merged.

PR #262, opened 2021-05-24
Support nested workflow_jobs

agrare:
When using nested workflows ansible tower returns a workflow_job's ID in the job_id property which causes the auto-lookup functionality of ansible_tower_ruby_client to query the wrong collection.

Fixes: https://github.com/ManageIQ/manageiq/issues/21228

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/automation_manager/workflow_job.rb
+++ b/app/models/manageiq/providers/ansible_tower/automation_manager/workflow_job.rb
@@ -97,7 +97,18 @@ class ManageIQ::Providers::AnsibleTower::AutomationManager::WorkflowJob <
     update_parameters(raw_workflow_job) if parameters.empty?
 
     raw_workflow_job.workflow_job_nodes.each do |node|
-      update_child_job(node.job) if node.job
+      next if node.job_id.nil?
+
+      case node.summary_fields&.job&.type
+      when "job"
+        update_child_job(node.job)
+      when "workflow_job"
+        # For nested workflows ansible tower returns a workflow_job id in the job
+        # attribute.  This means that "/api/v2/jobs/#{node.job_id}" will return a 404
+        # because the collection has to be "/api/v2/workflow_jobs/#{node.job_id}"
+        child_workflow_job = node.api.workflow_jobs.find(node.job_id)
+        update_child_workflow_job(child_workflow_job)
+      end
     end
   end
   private :update_with_provider_object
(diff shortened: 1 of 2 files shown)


agrare:
I need to update the mocks in the specs to work with the new check and add a test for nested workflows.

bdunne (approved):

mikebutak:
@agrare If I manually add these changes to our appliances, does that present any issues in the future when we upgrade to the next stable release?  We are still on Ivanchuk.  I wouldn't think so, but didn't want to assume.

agrare:
@mikebutak worst case you might upgrade to a version which doesn't have this patch backported and your change would be overwritten.

Fryguy:
Backported to `lasker` in commit 7332de2ab44250ce4067bb624e46e6d9109809fc.

```text
commit 7332de2ab44250ce4067bb624e46e6d9109809fc
Author: Brandon Dunne <bdunne@redhat.com>
Date:   Mon May 24 17:24:47 2021 -0400

    Merge pull request #262 from agrare/support_nested_workflow_job_templates
    
    Support nested workflow_jobs
    
    (cherry picked from commit dcbe5d04c942aa6fb305e5ee96fbb4b07aa651cd)
```

mikebutak:
@agrare @Fryguy after bringing in these changes, we get the following error when calling a workflow template:

```
 <AEMethod provision_if_selected> Template postgres soup to nuts for miq launching failed. Reason: undefined method `job_id' for #<AnsibleTowerClient::WorkflowJobNode:0x0000000013558620>
```

Note that this is not even a workflow template nested in a workflow template, just a "single-depth" workflow template.  

I'll add more logs for context:

```

agrare:
@mikebutak could you try changing `next if node.job_id.nil?` to `next !node.respond_to?(:job_id) || node.job_id.nil?`

agrare:
If that works I think I can implement this cleanly in the ansible_tower_client

mikebutak:
Will do @agrare !

mikebutak:
That seems to have addressed that error @agrare !  Looks like the single-depth workflow is working ok now.  Thank you

chessbyte:
@agrare probably need to get the fix @mikebutak validated into master and backported to Lasker

agrare:
@chessbyte :+1: yeah that's what I've been working on over [here](https://github.com/ansible/ansible_tower_client_ruby/pull/147)

mikebutak:
@agrare just FYI I also tested this with workflow templates that contain workflow templates and it works without errors.  Great news for our team.  Thank you!

mikebutak:
> @mikebutak could you try changing `next if node.job_id.nil?` to `next !node.respond_to?(:job_id) || node.job_id.nil?`

Hey @agrare I just tried to re-use these instructions in another environment.  I can't find `next if node.job_id.nil?` to replace it with your suggestion. And I can't find the suggested replacement text either.  Is this file already updated with the fix: https://github.com/ManageIQ/manageiq-providers-ansible_tower/blob/master/app/models/manageiq/providers/ansible_tower/automation_manager/workflow_job.rb?

chessbyte:
@mikebutak yes, that was done on master in this commit:  https://github.com/ManageIQ/manageiq-providers-ansible_tower/commit/8dfbf075c38387df426223bce273eac4c3085071#diff-02e7b535fe11bb1635fd971485e2e9b53ac8fedc9964e7182f6c8e599d57d4dc

PR #263, opened 2021-06-03
Check if workflow job node has a job_id

agrare:
Fix for the issue noted https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/262#issuecomment-853317690

Introduced by https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/262/files#diff-02e7b535fe11bb1635fd971485e2e9b53ac8fedc9964e7182f6c8e599d57d4dcR100

Not all workflow_job_nodes have a job_id

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/automation_manager/workflow_job.rb
+++ b/app/models/manageiq/providers/ansible_tower/automation_manager/workflow_job.rb
@@ -97,7 +97,7 @@ class ManageIQ::Providers::AnsibleTower::AutomationManager::WorkflowJob <
     update_parameters(raw_workflow_job) if parameters.empty?
 
     raw_workflow_job.workflow_job_nodes.each do |node|
-      next if node.job_id.nil?
+      next if node.try(:job_id).nil?
 
       case node.summary_fields&.job&.type
       when "job"
--- a/spec/models/manageiq/providers/ansible_tower/automation_manager/workflow_job_spec.rb
+++ b/spec/models/manageiq/providers/ansible_tower/automation_manager/workflow_job_spec.rb
@@ -139,6 +139,28 @@ describe ManageIQ::Providers::AnsibleTower::AutomationManager::WorkflowJob do
         expect(child_job.status).to                eq(the_raw_job.status)
         expect(child_job.ext_management_system).to eq(manager)
       end
+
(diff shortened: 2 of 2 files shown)


app/models/manageiq/providers/ansible_tower/automation_manager/workflow_job.rb
```diff
@@ -97,7 +97,7 @@ def update_with_provider_object(raw_workflow_job)
     update_parameters(raw_workflow_job) if parameters.empty?
 
     raw_workflow_job.workflow_job_nodes.each do |node|
-      next if node.job_id.nil?
+      next if !node.respond_to?(:job_id) || node.job_id.nil?
```
chessbyte, line 100: would the following read better and do the same thing?

`next if node&.job_id.nil?`

agrare: Yes, I was perhaps over-cautiously trying to use exactly what the ansible_tower_client uses: https://github.com/ansible/ansible_tower_client_ruby/blob/master/lib/ansible_tower_client/base_models/workflow_job_node.rb#L13

agrare: Also `&.` doesn't handle method_missing like `.try()` does, so it would have to be `node.try(:job_id)`

chessbyte: gotcha - makes sense

agrare: Added a spec to cover the method_missing case, feel more comfortable with `.try()` now

Fryguy:
Backported to `lasker` in commit d5ec9817e49c1be6b4a153bc8b70701d89c07f45.

```text
commit d5ec9817e49c1be6b4a153bc8b70701d89c07f45
Author: Oleg Barenboim <chessbyte@gmail.com>
Date:   Thu Jun 3 14:34:50 2021 -0400

    Merge pull request #263 from agrare/fix_workflows_without_job_id
    
    Check if workflow job node has a job_id
    
    (cherry picked from commit 852d644696053afc0a7d9c2d2e864a8c3c97694a)
```

PR #267, opened 2021-10-14
Fix merge_extra_vars with nil variables

agrare:
Fix ConfigurationScript/ConfigurationWorkflow merge_extra_vars with `nil` variables


Fryguy:
Do we know how `variables` can become nil?

PR #272, opened 2022-01-13
Use cross-link vms collection from core

agrare:
Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/21654

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/inventory/parser/automation_manager.rb
+++ b/app/models/manageiq/providers/ansible_tower/inventory/parser/automation_manager.rb
@@ -23,7 +23,7 @@ class ManageIQ::Providers::AnsibleTower::Inventory::Parser::AutomationManager < 
       inventory_object.hostname = host.name
       inventory_object.virtual_instance_ref = host.instance_id
       inventory_object.inventory_root_group = persister.inventory_root_groups.lazy_find(host.inventory_id.to_s)
-      inventory_object.counterpart = persister.vms.lazy_find(host.instance_id)
+      inventory_object.counterpart = persister.cross_link_vms.lazy_find({:uid_ems => host.instance_id}, {:ref => :by_uid_ems})
     end
   end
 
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/ansible_tower/inventory/persister/definitions/automation_collections.rb
```diff
@@ -32,14 +34,4 @@ def add_configuration_script_payloads(extra_properties = {})
       )
     end
   end
-
-  def add_vms
```
chessbyte, line 36: where was `add_vms` being called from?  don't seem to see that caller being deleted in this PR

agrare: It is [[here]](https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/272/files#diff-4949e15a250d6715677f53bc2c31a7e36a634a7b85529019198fe64ec1b78574R20) on line 20 just forgot to remove it thanks

app/models/manageiq/providers/ansible_tower/inventory/parser/automation_manager.rb
```diff
@@ -23,7 +23,7 @@ def configured_systems
       inventory_object.hostname = host.name
       inventory_object.virtual_instance_ref = host.instance_id
       inventory_object.inventory_root_group = persister.inventory_root_groups.lazy_find(host.inventory_id.to_s)
-      inventory_object.counterpart = persister.vms.lazy_find(host.instance_id)
+      inventory_object.counterpart = persister.cross_link_vms.lazy_find({:uid_ems => host.instance_id}, {:ref => :by_uid_ems})
```
agrare, line 26: NOTE this was using the uid_ems as the manager_ref previously due to the vms collection definition in the automation persister: https://github.com/ManageIQ/manageiq/pull/21654/files#diff-7422cd6fb687662faa3ce1db090a4888c9ef4c71b952e08d94435bd115d0d6b4L47

PR #274, opened 2022-01-31
add supports :create for scripts

kbrock:
requires

- [x] ManageIQ/manageiq#21674

to get away from `respond_to?(:create)` kind of behavior,
adding `supports :create` to:

- Credential
- ConfigurationScriptSource
- ConfigurationScript

this has been a trend and was brought up in https://github.com/ManageIQ/manageiq-api/pull/1116

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/automation_manager/configuration_script.rb
+++ b/app/models/manageiq/providers/ansible_tower/automation_manager/configuration_script.rb
@@ -3,6 +3,8 @@ class ManageIQ::Providers::AnsibleTower::AutomationManager::ConfigurationScript 
   include ProviderObjectMixin
   include ManageIQ::Providers::AnsibleTower::AutomationManager::TowerApi
 
+  supports :create
+
   def run_with_miq_job(options, userid = nil)
     options[:name] = "Job Template: #{name}"
     options[:ansible_template_id] = id
--- a/app/models/manageiq/providers/ansible_tower/automation_manager/configuration_script_source.rb
+++ b/app/models/manageiq/providers/ansible_tower/automation_manager/configuration_script_source.rb
@@ -4,6 +4,8 @@ class ManageIQ::Providers::AnsibleTower::AutomationManager::ConfigurationScriptS
   include ManageIQ::Providers::AnsibleTower::AutomationManager::TowerApi
   include ProviderObjectMixin
 
+  supports :create
+
   def self.provider_params(params)
     if params.keys.include?(:authentication_id)
       authentication_id = params.delete(:authentication_id)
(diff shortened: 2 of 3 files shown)


app/models/manageiq/providers/ansible_tower/automation_manager/tower_api.rb
```diff
@@ -1,6 +1,11 @@
 module ManageIQ::Providers::AnsibleTower::AutomationManager::TowerApi
   extend ActiveSupport::Concern
 
+  def self.included(model)
+    model.include SupportsFeatureMixin
+    model.supports :create
+  end
```
Fryguy, line 7: We have ActiveSupport::Concern, so we should do:

```suggestion
  included do
    include SupportsFeatureMixin
    supports :create
  end
```

app/models/manageiq/providers/ansible_tower/automation_manager/tower_api.rb
```diff
@@ -1,6 +1,11 @@
 module ManageIQ::Providers::AnsibleTower::AutomationManager::TowerApi
   extend ActiveSupport::Concern
 
+  def self.included(model)
+    model.include SupportsFeatureMixin
+    model.supports :create
```
agrare, line 6: Good with this module defining `supports :create` because the module actually implements `raw_create_in_provider` down on [[line 10]](https://github.com/ManageIQ/manageiq-providers-ansible_tower/pull/274/files#diff-2f371a29d9bf166fc5c677c504445d485bb3cbf53497b922f90f0dfa6a8fdd32R10) (re: https://github.com/ManageIQ/manageiq/pull/21674/files#r796072565)

kbrock: yea, I'm getting cold feet on this one.

think we don't mixin `supports :create` anywhere else. thinking no reason to start now

kbrock:
update:
moved `supports :create` to the individual models

app/models/manageiq/providers/ansible_tower/automation_manager/configuration_script.rb
```diff
@@ -1,7 +1,11 @@
 class ManageIQ::Providers::AnsibleTower::AutomationManager::ConfigurationScript <
   ManageIQ::Providers::ExternalAutomationManager::ConfigurationScript
+
   include ProviderObjectMixin
   include ManageIQ::Providers::AnsibleTower::AutomationManager::TowerApi
+  include SupportsFeatureMixin
```
agrare, line 6: This is usually only done in core in the base model

kbrock: looks like it is not defined in master yet.

kbrock:
update:

- moved `includes` to manageiq (dependent PR)

agrare (approved):

PR #278, opened 2022-02-09
Replace .parent with .module_parent

agrare:
`DEPRECATION WARNING: 'Module#parent' has been renamed to 'module_parent'. 'parent' is deprecated and will be removed in Rails 6.1.`


jrafanie:
❤️ LGTM, will merge when green

PR #305, opened 2024-02-26
[RADJABOV] Test radjabov with ruby 3.1

jrafanie:
Test radjabov with ruby 3.1


jrafanie:
This is failing due to the change in core: https://github.com/ManageIQ/manageiq/commit/5b5e9273efafc8954db11cc160306cea0d13681c

I think it's safe to merge and we'll probably revert that commit unless we find a good solution.

PR #306, opened 2024-02-28
Workaround attribute default triggering changes

jrafanie:
This was caused by the change to use attribute default in enabled on ExtManagementSystem from this commit:
https://github.com/ManageIQ/manageiq/commit/5b5e9273efafc8954db11cc160306cea0d13681c

Same change as: https://github.com/ManageIQ/manageiq-providers-awx/pull/27


jrafanie:
Closing.  Will be reverting https://github.com/ManageIQ/manageiq/commit/5b5e9273efafc8954db11cc160306cea0d13681c and the whole PR: https://github.com/ManageIQ/manageiq/pull/22901

Will be looking to use database defaults, which don't have the bad side effect of defaults triggering "changed", for things that have static values and don't have different vlues in STI classes.

PR #318, opened 2024-12-11
Rename 'Ansible Tower' to 'Ansible Automation Platform'

nasark:
@miq-bot assign @agrare
@miq-bot add_reviewer @agrare

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/automation_manager/configuration_script.rb
+++ b/app/models/manageiq/providers/ansible_tower/automation_manager/configuration_script.rb
@@ -15,8 +15,8 @@ class ManageIQ::Providers::AnsibleTower::AutomationManager::ConfigurationScript 
   end
 
   def self.display_name(number = 1)
-    n_('Job Template (Ansible Tower)', 'Job Templates (Ansible Tower)', number)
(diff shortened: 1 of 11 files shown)


app/models/manageiq/providers/ansible_tower/automation_manager/configuration_script.rb
```diff
@@ -15,8 +15,8 @@ def run_with_miq_job(options, userid = nil)
   end
 
   def self.display_name(number = 1)
-    n_('Job Template (Ansible Tower)', 'Job Templates (Ansible Tower)', number)
+    n_('Job Template (Ansible Automation Platform)', 'Job Templates (Ansible Automation Platform)', number)
```
agrare, line 18: For a follow-up but I wonder how many of these can be dropped if we just did 
```suggestion
    n_("Job Template (#{module_parent.description})", "Job Templates (#{module_parent.description})", number)
```
in the base AWS class

Fryguy: Because there's the `n_`, you can't use interpolation, but you could do:

```suggestion
    n_('Job Template (%{provider_description}', 'Job Templates (%{provider_description)', number % {:provider_description => module_parent.description}
```

PR #322, opened 2025-02-26
Manually define aliased methods

jrafanie:
Fixes a deprecation warning on non-attributes:

```
DEPRECATION WARNING: ManageIQ::Providers::AnsibleTower::AutomationManager::Credential model aliases `resource`, but `resource` is not an attribute. Starting in Rails 7.2, alias_attribute with non-attribute targets will raise. Use `alias_method :manager, :resource` or define the method manually. (called from new at /home/runner/work/manageiq/manageiq/app/models/mixins/new_with_type_sti_mixin.rb:17)
```

See also: https://github.com/ManageIQ/manageiq/commit/c76f6527da73bdad00883ae3f0fe4f5d907876a2

### DIFF
--- a/app/models/manageiq/providers/ansible_tower/automation_manager/credential.rb
+++ b/app/models/manageiq/providers/ansible_tower/automation_manager/credential.rb
@@ -6,9 +6,16 @@ class ManageIQ::Providers::AnsibleTower::AutomationManager::Credential < ManageI
   # CUD operations in the TowerApi concern
 
   alias_attribute :manager_id, :resource_id
-  alias_attribute :manager, :resource
 
   supports :create
 
   FRIENDLY_NAME = 'Ansible Automation Platform Credential'.freeze
+
+  def manager
+    resource
+  end
+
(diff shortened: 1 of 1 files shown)


agrare:
@jrafanie it looks like in https://github.com/ManageIQ/manageiq-providers-awx/pull/37 you did this a different way, could we delete these entirely since we're inheriting from `ManageIQ::Providers::Awx::AutomationManager::Credential` ?

jrafanie:
> @jrafanie it looks like in [ManageIQ/manageiq-providers-awx#37](https://github.com/ManageIQ/manageiq-providers-awx/pull/37) you did this a different way, could we delete these entirely since we're inheriting from `ManageIQ::Providers::Awx::AutomationManager::Credential` ?

yeah, good call.  I'll check.

jrafanie:
Closing as the branch name is now wrong...will delete this code instead.  ❤️

jrafanie:
replaced by #323

PR #338, opened 2026-02-05
Validate URL format for Ansible Tower URL

agrare:
Depends:
- [x] https://github.com/ManageIQ/manageiq/pull/23719
- [x] https://github.com/ManageIQ/manageiq-providers-awx/pull/50

Related:
* https://github.com/ManageIQ/manageiq-providers-awx/pull/50


app/models/manageiq/providers/ansible_tower/provider.rb
```diff
@@ -1,182 +1,8 @@
...
-
-    verify_connection(raw_connect(url, userid, password, verify_ssl))
-  end
-
-  def self.default_api_path
-    "/api/v2".freeze
-  end
-
-  def self.adjust_url(url)
-    url = "https://#{url}" unless url =~ %r{\Ahttps?:\/\/} # HACK: URI can't properly parse a URL with no scheme
```
agrare, line 114: I just noticed this was a 99.9% copy of `Awx::Provider` but this defaults to `https` instead of `http` so I'll have to update the schema data migration

agrare: https://github.com/ManageIQ/manageiq-schema/pull/831

app/models/manageiq/providers/ansible_tower/provider.rb
```diff
@@ -1,182 +1,8 @@
...
-
-  def ensure_managers
-    build_automation_manager unless automation_manager
-    automation_manager.provider = self
-
-    if zone_id_changed?
-      automation_manager.zone    = zone
-      automation_manager.enabled = Zone.maintenance_zone&.id != zone_id
-    end
-  end
```
agrare, line 181: The Awx::Provider file was almost a complete copy of this so we can inherit instead of copying it

agrare:
@miq-bot cross-repo-test https://github.com/ManageIQ/manageiq/pull/23719

agrare:
Because we're inheriting from `Awx::Provider` this actually depends on https://github.com/ManageIQ/manageiq-providers-awx/pull/50 to pass the new specs

agrare:
@miq-bot cross-repo-tests https://github.com/ManageIQ/manageiq-providers-awx/pull/50

agrare:
The bot is a little slow today

agrare:
Cross-repo tests green: https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/1029

